PR #6, opened 2021-09-23
Add ability to restart failed build items

anfimovdm:


alws/crud.py
```diff
@@ -186,6 +187,15 @@ async def get_available_build_task(
     return db_task
 
 
+async def update_failed_build_items(db: Session, build_id: int):
+    query = sqlalchemy.and_(models.BuildTask.build_id == build_id, 
+                            models.BuildTask.status == BuildTaskStatus.FAILED)
+    async with db.begin():
+        await db.execute(update(models.BuildTask).where(query).values(
+            status=BuildTaskStatus.IDLE))
+        await db.commit()
```
AvoidMe, line 196: Please, also include right build order for restarted items, you can check `alws/build_planner.py`
Right now, if you have two failed items in your build, after restart they will be run together

AvoidMe (changes requested):

alws/crud.py
```diff
@@ -23,7 +23,8 @@
 from alws.utils.distro_utils import create_empty_repo
 
 
-__all__ = ['create_build', 'get_builds', 'create_platform', 'get_platforms']
+__all__ = ['create_build', 'get_builds', 'create_platform',
```
soksanichenko, line 26: FYI. This decoration of the code is better than current because you can add/remove/comment each element with no additional efforts by an one hotkey)

```
lst = [
  'element1',
  'element2',
  'element3',
]
```

PR #12, opened 2021-09-28
Sign node api

oweel:


alws/constants.py
```diff
@@ -7,11 +7,18 @@
...
-    STARTED = 1
-    COMPLETED = 2
-    FAILED = 3
-    EXCLUDED = 4
+    BUILD_STARTED = 1
+    BUILD_COMPLETED = 2
+    SIGN_STARTED = 3
+    SIGN_COMPLETED = 4
+    FAILED = 5
+    EXCLUDED = 6
```
AvoidMe, line 15: Please, make a separate status for signing, just like we did for testing

alws/crud.py
```diff
@@ -181,7 +181,44 @@ async def get_available_build_task(
...
+        ) -> models.BuildTask:
+    async with db.begin():
+        ts_expired = datetime.datetime.now() - datetime.timedelta(minutes=20)
+        query = ~models.BuildTask.dependencies.any()
+        db_task = await db.execute(
+            select(models.BuildTask).where(query).with_for_update().filter(
+                    sqlalchemy.and_(
+                        models.BuildTask.status < BuildTaskStatus.SIGN_COMPLETED,
+                        models.BuildTask.status >= BuildTaskStatus.BUILD_COMPLETED,
+                        models.BuildTask.arch.in_(request.supported_arches),
```
AvoidMe, line 201: rpmsign doesn't require from rpm to be same arch as host machine, so, `supported_arches` can be removed

alws/crud.py
```diff
@@ -275,9 +366,9 @@ async def github_login(
             settings.github_client_secret
         )
         github_info = await get_github_user_info(github_user_token)
-        if not any(item for item in github_info['organizations']
-                   if item['login'] == 'AlmaLinux'):
-            return
+        # if not any(item for item in github_info['organizations']
+                #    if item['login'] == 'AlmaLinux'):
+            # return
```
AvoidMe, line 371: Issue with login was fixed, please, remove this.

alws/routers/sign_node.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+@router.post('/ping')
+async def ping(
+            node_status: build_task_schema.Ping,
+            db: database.Session = Depends(get_db)
+        ):
+    if not node_status.active_tasks:
+        return
+    await crud.ping_tasks(db, node_status.active_tasks)
```
AvoidMe, line 24: Can be safely removed

alws/models.py
```diff
@@ -163,11 +163,13 @@ class BuildTask(Base):
...
         nullable=False
     )
+    pgp_key_id = sqlalchemy.Column(sqlalchemy.VARCHAR(length=50), nullable=False)
     status = sqlalchemy.Column(sqlalchemy.Integer, nullable=False)
     index = sqlalchemy.Column(sqlalchemy.Integer, nullable=False)
     arch = sqlalchemy.Column(sqlalchemy.VARCHAR(length=50), nullable=False)
     ref = relationship('BuildTaskRef')
-    artifacts = relationship('BuildTaskArtifact')
+    artifacts = relationship('BuildTaskArtifact', backref='artifacts')
+    signed_artifacts = relationship("BuildTaskArtifact", backref='signed_artifacts')
```
AvoidMe, line 172: I think, it's better to move `pgp_key_id` to `Build` model, also, instead of `signed_artifacts` we can just replace links for unsigned one, since I doubt what we will store both of them.

AvoidMe (changes requested):

AvoidMe:
Also, please, rebase your pull-request!

alws/crud.py
```diff
@@ -250,9 +250,29 @@ async def modify_distribution(build_id: int, distribution: str, db: Session,
             await pulp_client.modify_repository(remove=value, repo_to=key)
 
 
+async def sign_start(
+            db: Session, 
+            build_task_id: int
+        ) -> models.BuildTask:
+    async with db.begin():
+        ts_expired = datetime.datetime.now() - datetime.timedelta(minutes=20)
+        query = models.BuildTask.id == build_task_id
```
anfimovdm, line 259: I think, that doesn\`t look like query, may be `where_condition` or something else?
Also, you can assign query in variable then use it in db.execute()
```python
query = select(...).where(...).with_for_update()
db_task = await db.execute(query)
```

alws/crud.py
```diff
@@ -476,9 +589,9 @@ async def github_login(
             settings.github_client_secret
         )
         github_info = await get_github_user_info(github_user_token)
-        if not any(item for item in github_info['organizations']
-                   if item['login'] == 'AlmaLinux'):
-            return
+        # if not any(item for item in github_info['organizations']
+        #            if item['login'] == 'AlmaLinux'):
+        #     return
```
anfimovdm, line 594: It's mistake?

alws/alembic/versions/2fb5ffc7ff29_add_pgp_key_id.py
```diff
@@ -0,0 +1,29 @@
...
+down_revision = '76eec955d203'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column(
+        "builds",
+        sa.Column(
+            "pgp_key_id", sa.Text(), nullable=False
```
rpisarev, line 23: Why it cannot be nullable? There will be some builds non-signed yet. Or defaults is an empty string?

alws/constants.py
```diff
@@ -17,6 +17,20 @@ def is_finished(cls, status):
...
+    IDLE = 0
+    PENDING = 1
+    STARTED = 2
+    COMPLETED = 3
+    FAILED = 4
+    EXCLUDED = 5
+
+    @classmethod
+    def is_finished(cls, status):
+        return status not in (cls.IDLE, cls.COMPLETED)
```
rpisarev, line 31: I guess FAILED is also a finish in some way. Are you sure we need EXCLUDED ? I don't catch the case.

alws/crud.py
```diff
@@ -290,6 +310,42 @@ async def get_available_build_task(
     return db_task
 
 
+async def get_available_sign_task(
+            db: Session, 
+            request: build_task_schema.RequestTask
+        ) -> models.BuildTask:
+    async with db.begin():
+        ts_expired = datetime.datetime.now() - datetime.timedelta(minutes=20)
+        query = ~models.BuildTask.dependencies.any()
```
rpisarev, line 319: Does it mean the signing process works as a build process with dependency?

alws/models.py
```diff
@@ -194,6 +194,7 @@ class Build(Base):
         secondaryjoin=(BuildDependency.c.build_dependency == id)
     )
     mock_options = sqlalchemy.Column(JSONB)
+    pgp_key_id = sqlalchemy.Column(sqlalchemy.VARCHAR(length=50), nullable=False)
```
rpisarev, line 197: Here you say about VARCHAR, but in alembic migration, the field is a Text. Is it OK? There different types for DB.

alws/alembic/versions/2fb5ffc7ff29_add_pgp_key_id.py
Korulag: More than that, all our builds are not signed at the beginning

alws/constants.py
Korulag: EXCLUDED is not needed for sign because we will not sign anything for excluded items

alws/crud.py
oweel: This query works. I checked it. Or I misunderstood you.

alws/crud.py
```diff
@@ -370,6 +430,63 @@ async def build_done(
         await db.commit()
 
 
+async def sign_done(
```
anfimovdm, line 433: Same logic as in `albs-web-server.alws.crud.build_done`. Does that really necessary?

anfimovdm (changes requested):

PR #28, opened 2021-10-29
Added functionality to bootstrap production repos.

Korulag:
This commit brings the ability to bootstrap production repositories in Pulp to be able to release into them later.
Tickets: ALBS:18, ALBS-29


alws/app.py
```diff
@@ -37,5 +38,5 @@ async def shutdown():
 
 
 for module in (builds, build_node, distro, platforms, projects, releases,
-               tests, users):
+               repositories, tests, users):
```
soksanichenko, line 41: Why did you not move it to some external constant which will be imported and used here?

alws/crud.py
```diff
@@ -782,3 +788,228 @@ async def commit_release(db: Session, release_id: int) -> (models.Release, str):
         selectinload(models.Release.platform)
     ))
     return release_res.scalars().first(), message
+
+
+async def get_repositories(db: Session, repository_id: int = None
+                           ) -> typing.List[models.Repository]:
+    repo_q = select(models.Repository)
+    if repository_id:
```
soksanichenko, line 796: The Zen of Python: Explicit is better than implicit.
So, `repository_id is not None` is better. And it helps to catch errors in the case when repository_id is zero or empty string.

alws/crud.py
```diff
@@ -782,3 +788,228 @@ async def commit_release(db: Session, release_id: int) -> (models.Release, str):
...
+    return repository
+
+
+async def update_repository(
+        db: Session, repository_id: int,
+        payload: repository_schema.RepositoryUpdate
+) -> models.Repository:
+    async with db.begin():
+        result = await db.execute(select(models.Repository).where(
+            models.Repository.id == repository_id))
```
soksanichenko, line 870: Why do you not use `get(repository_id`?

soksanichenko (changes requested):

alws/crud.py
Korulag: https://stackoverflow.com/a/51997031 in this particular case you're right, but it's not always a correct way. I'll fix it

alws/crud.py
```diff
@@ -782,3 +788,235 @@ async def commit_release(db: Session, release_id: int) -> (models.Release, str):
...
+
+
+async def update_repository(
+        db: Session, repository_id: int,
+        payload: repository_schema.RepositoryUpdate
+) -> models.Repository:
+    async with db.begin():
+        repository = await db.execute(select(
+            models.Repository).get(repository_id))
+        # repository = result.scalars().first()
```
anfimovdm, line 889: Is that mistake?

anfimovdm (approved):

anfimovdm (approved):

alws/crud.py
```diff
@@ -867,3 +874,223 @@ async def commit_release(db: Session, release_id: int) -> (models.Release, str):
...
+            models.Repository.id == repository_id))
+        await db.commit()
+
+
+async def add_to_platform(db: Session, platform_id: int,
+                          repository_ids: typing.List[int]) -> models.Platform:
+    platform_result = await db.execute(select(models.Platform).where(
+        models.Platform.id == platform_id).options(
+        selectinload(models.Platform.repos)).with_for_update())
+    platform = platform_result.scalars().first()
```
rpisarev, line 995: Platform could be not found in general case

alws/routers/platforms.py
```diff
@@ -33,3 +33,19 @@ async def modify_platform(
...
+              response_model=platform_schema.Platform)
+async def add_repositories_to_platform(platform_id: int,
+                                       repositories_ids: typing.List[int],
+                                       db: database.Session = Depends(get_db)):
+    return await crud.add_to_platform(db, platform_id, repositories_ids)
+
+
+@router.patch('/{platform_id}/remove-repositories',
+              response_model=platform_schema.Platform)
+async def add_repositories_to_platform(platform_id: int,
```
rpisarev, line 48: It seems you miss renaming the function

Korulag: Thanks, fixed

alws/crud.py
Korulag: Yeah, you're right, it may be, will fix

JohnKepplers (approved):

rpisarev (approved):

PR #30, opened 2021-11-07
ALBS-20: Added flag whether "oracle" is enabled.

Korulag:
We may have a situation when the system that is predicting packages is
unavailable and we need to just create a release plan on our own. This
commit brings support for such case.

### DIFF
--- a/alws/releases.py
+++ b/alws/releases.py
@@ -62,18 +62,10 @@ class OracleClient:
                 return json
 
 
-async def get_release_plan(db: Session, build_ids: typing.List[int],
-                           base_dist_name: str, base_dist_version: str,
-                           reference_dist_name: str,
-                           reference_dist_version: str) -> dict:
-    # FIXME: put actual endpoint to use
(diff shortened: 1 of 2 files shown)


alws/releases.py
```diff
@@ -83,16 +75,19 @@ async def get_release_plan(db: Session, build_ids: typing.List[int],
...
-                    models.SourceRpm.artifact),
-            selectinload(
-                models.Build.binary_rpms).selectinload(
-                    models.BinaryRpm.artifact)
+        selectinload(
+            models.Build.source_rpms).selectinload(
+            models.SourceRpm.artifact),
+        selectinload(
+            models.Build.binary_rpms).selectinload(
+            models.BinaryRpm.artifact)
```
anfimovdm, line 83: I'm think there is a wrong indents

anfimovdm (approved):

rpisarev (commented):
lgtm

JohnKepplers (approved):

PR #43, opened 2021-11-17
ALBS-67: Added ref type for build tasks

anfimovdm:
- Added column `ref_type` in `BuildTaskRef` model
- Added new `alembic` migration
- Added new enum `BuildTaskRefType`
- Added `ref_type` in `build_schema.BuildTaskRef`


rpisarev (commented):
lgtm

alws/alembic/versions/585ab06efb32_extended_build_task_refs.py
```diff
@@ -0,0 +1,30 @@
...
+revision = '585ab06efb32'
+down_revision = 'cbd577117a13'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('build_task_refs', sa.Column('ref_type', sa.Integer(), nullable=True))
+    op.create_unique_constraint(None, 'repositories', ['name', 'arch', 'type', 'debug'])
```
Korulag, line 22: Add constraint name, because otherwise it will fail to be dropped in case someone needs downgrade

Korulag (changes requested):

alws/alembic/versions/585ab06efb32_extended_build_task_refs.py
anfimovdm: Thanks, fixed

Korulag (approved):

PR #52, opened 2021-12-09
ALBS-84 Repository ID for release response. Beholder communication

monkrobot:
Add repository ID to response from backend for release. Fixed communication with beholder service

### DIFF
--- a/alws/releases.py
+++ b/alws/releases.py
@@ -1,16 +1,14 @@
 import re
 import typing
-import urllib.parse
 from collections import namedtuple
 
-import aiohttp
 import jmespath
-import json
 from sqlalchemy import select
 from sqlalchemy.orm import Session, selectinload
 
 from alws import models
 from alws.config import settings
+from alws.utils.beholder_client import BeholderClient
 from alws.utils.pulp_client import PulpClient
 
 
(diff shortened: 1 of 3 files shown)


alws/releases.py
```diff
@@ -110,8 +79,11 @@ async def get_release_plan(db: Session, build_ids: typing.List[int],
                            base_dist_name: str, base_dist_version: str,
                            reference_dist_name: str,
                            reference_dist_version: str) -> dict:
-    # FIXME: put actual endpoint to use
-    endpoint = f'/api/v1/distros/{reference_dist_name}/' \
+    clean_ref_dist_name = re.search(
+        r'(?P<dist_name>[a-z]+)', reference_dist_name,
+        re.IGNORECASE).groupdict().get('dist_name')
+    clean_ref_dist_name_lower = clean_ref_dist_name.lower()
```
soksanichenko, line 85: Can var `clean_ref_dist_name` be None?

alws/releases.py
```diff
@@ -110,8 +79,11 @@ async def get_release_plan(db: Session, build_ids: typing.List[int],
...
+    endpoint = f'/api/v1/distros/{clean_ref_dist_name}/' \
                f'{reference_dist_version}/projects/'
     packages = []
     repo_name_regex = re.compile(r'\w+-\d-(?P<name>\w+(-\w+)?)')
```
soksanichenko, line 89: Module `re` compiles and caches regexps inside itself, so you don't need do it. FYI

alws/utils/beholder_client.py
```diff
@@ -23,6 +25,15 @@ async def get(self, endpoint: str,
         full_url = self._get_url(endpoint)
         async with aiohttp.ClientSession(headers=req_headers) as session:
             async with session.get(full_url, params=params) as response:
-                json = await response.json()
+                json_ = await response.json()
```
soksanichenko, line 28: bad name of the var. Some `json_dict` or `json_data` is better

alws/utils/beholder_client.py
```diff
@@ -23,6 +25,15 @@ async def get(self, endpoint: str,
...
                 response.raise_for_status()
```
soksanichenko, line 29: You should call `raise_for_status()` before reading any data. And I'm not sure that aiohttp has the same method

soksanichenko (changes requested):

alws/releases.py
```diff
@@ -121,23 +93,26 @@ async def get_release_plan(db: Session, build_ids: typing.List[int],
...
+
     if not settings.package_oracle_enabled:
         return {
             'packages': [{'package': pkg, 'repositories': []}
                          for pkg in pulp_packages],
             'repositories': prod_repos
         }
 
-    oracle_response = await OracleClient(settings.packages_oracle_host).post(
+    oracle_response = await BeholderClient(settings.packages_oracle_host).post(
```
anfimovdm, line 114: If you deleted OracleClient, so there doesn’t any sense to use packages_oracle_host, better to use var for beholder host

anfimovdm (changes requested):

alws/releases.py
monkrobot: No, it can't be None

alws/releases.py
monkrobot: Thanks for info :)

alws/utils/beholder_client.py
```diff
@@ -23,6 +25,16 @@ async def get(self, endpoint: str,
         full_url = self._get_url(endpoint)
         async with aiohttp.ClientSession(headers=req_headers) as session:
             async with session.get(full_url, params=params) as response:
-                json = await response.json()
                 response.raise_for_status()
```
soksanichenko, line 28: https://docs.aiohttp.org/en/stable/client_reference.html, look at param `raise_for_status`

soksanichenko (changes requested):

anfimovdm (approved):

soksanichenko (approved):

PR #62, opened 2021-12-23
ALBS-106: Fix release plan generation

Korulag:
This commit brings fixes for release plan generation in different cases
like when beholder is not available or has no information on how
packages should be put into the repositories.


alws/routers/releases.py
```diff
@@ -21,11 +21,11 @@ async def get_releases(db: database.Session = Depends(get_db)):
 
 
 @router.post('/new/', response_model=release_schema.Release)
-async def create_new_release(builds: release_schema.ReleaseCreate,
+async def create_new_release(payload: release_schema.ReleaseCreate,
                              db: database.Session = Depends(get_db),
                              user: dict = Depends(JWTBearer())):
     release = await r_crud.create_new_release(
-        db, user['identity']['user_id'], builds)
+        db, user['identity']['user_id'], payload)
```
rpisarev, line 28: Why do you prefer using abstract word instead of more precise?

rpisarev (commented):
lgtm

anfimovdm (approved):

monkrobot (approved):

alws/routers/releases.py
Korulag: Not sure actually, will think how to make names more specific

PR #64, opened 2021-12-27
ALBS-28. Dist macro should be automatically taken from a git tag.

rpisarev:
ALBS-28. Dist macro should be automatically taken from a git tag.
But if 'definitions.dist' is set this value will be used  instead of any parsed

### DIFF
--- a/alws/build_planner.py
+++ b/alws/build_planner.py
@@ -260,6 +260,8 @@ class BuildPlanner:
             mock_options: typing.Optional[dict[str, typing.Any]] = None,
             ref_platform_version: typing.Optional[str] = None):
         parsed_dist_macro = parse_git_ref('(el[\d]+_[\d]+)', ref.git_ref)
+        if mock_options is None:
+            mock_options = {'definitions': {}}
         dist_taken_by_user = mock_options['definitions'].get('dist', False)
         for platform in self._platforms:
             arch_tasks = []
--- a/alws/build_planner.py
+++ b/alws/build_planner.py
@@ -14,6 +14,7 @@ from alws.config import settings
 from alws.schemas import build_schema
 from alws.constants import BuildTaskStatus, BuildTaskRefType
 from alws.utils.pulp_client import PulpClient
+from alws.utils.parsing import parse_git_ref
 from alws.utils.modularity import ModuleWrapper, calc_dist_macro
 from alws.utils.gitea import download_modules_yaml, GiteaClient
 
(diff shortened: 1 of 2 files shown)


alws/build_planner.py
```diff
@@ -282,7 +285,10 @@ async def _add_single_ref(
                         build_index,
                         platform_dist
                     )
-                    mock_options['definitions']['dist'] = dist_macro
+                    if not dist_taken_by_user:
+                        mock_options['definitions']['dist'] = dist_macro
+                if not dist_taken_by_user and parsed_dist_macro:
+                    mock_options['definitions']['dist'] = parsed_dist_macro
```
soksanichenko, line 291: Should we catch the case when dist_macro is None?

rpisarev: Do you mean None from the user side (UI, API etc) ? Or automatically parsing for modules? (In case when it is not module build we catch the case)

rpisarev: Do you mean None from the user side (UI, API etc) ? Or automatically parsing for modules? (In case when it is not module build we catch the case)

soksanichenko: I mean the 2nd case

rpisarev: As I see the code of calc_dist_macro() it cannot return None

soksanichenko (approved):

anfimovdm (changes requested):

alws/build_planner.py
```diff
@@ -258,6 +259,8 @@ async def _add_single_ref(
             ref: models.BuildTaskRef,
             mock_options: typing.Optional[dict[str, typing.Any]] = None,
             ref_platform_version: typing.Optional[str] = None):
+        parsed_dist_macro = parse_git_ref('(el[\d]+_[\d]+)', ref.git_ref)
+        dist_taken_by_user = mock_options['definitions'].get('dist', False)
```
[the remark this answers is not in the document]
rpisarev: Maybe. It's the current behavior. Should I change it?

alws/utils/parsing.py
```diff
@@ -0,0 +1,12 @@
+import re
+
+
+__all__ = ['parse_git_ref']
+
+def parse_git_ref(pattern : str, git_ref : str):
```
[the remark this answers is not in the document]
rpisarev: OK, I will fix it

alws/build_planner.py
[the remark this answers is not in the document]
anfimovdm: I'm think you need to check this before changing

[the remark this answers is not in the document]
rpisarev: Fixed

anfimovdm (approved):

soksanichenko (approved):

monkrobot (approved):

Korulag (approved):

PR #71, opened 2022-01-12
Add almalinux-9 platform

eabdullin1:
We need to build almalinux-9-beta, so I tried to add a platform
I'm not sure that all is right.

### DIFF
--- a/reference_data/platforms.yaml
+++ b/reference_data/platforms.yaml
@@ -1,5 +1,5 @@
 ---
-  name: AlmaLinux-8
+- name: AlmaLinux-8
   distr_type: rhel
   distr_version: '8'
   test_dist_name: almalinux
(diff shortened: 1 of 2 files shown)


reference_data/platforms.yaml
```diff
@@ -201,3 +201,180 @@
...
+    versions:
+      - name: '9'
+        version_prefix: '90000'
+        dist_prefix: 'el9.0.0'
+  type: rpm
+  arch_list:
+    - x86_64
+    - aarch64
+    - ppc64le
+    - s390
```
Korulag, line 224: 2 things:
1. We need i686 arch anyway since we have multilib
2. I think `s390` is called `s390x`

Korulag (changes requested):

Korulag (approved):

PR #74, opened 2022-01-17
Devel modules can be built

AvoidMe:
This commit allows building devel-modules alongside with default ones.
Whenever you're building module now, we'll check for it's devel version and build it together, if present

There are significant api changes in `alws/utils/modularity.py`:
* `class ModuleWrapper` now contains only one stream, without module/index (Initial implementation was backported from old bs and required changes to build multiple modules in one modules.yaml). You can't make new ModuleWrapper from string anymore if template has more than one module in it.
* `class IndexWrapper` (new class) is now should be used if you don't know for sure, how many module templates inside modules.yaml, or want to build multiple modules. All places in code, where we can find multiple modules at once, now using `IndexWrapper` instead of `ModuleWrapper`

`PulpClient.create_module` now requires name/stream/context/arch, while before it was trying to parse template with `ModuleWrapper` and figured out settings by itself, but since single template can have more than one module now, we're leaving this params for user to define.

### DIFF
--- a/alws/build_planner.py
+++ b/alws/build_planner.py
@@ -1,9 +1,9 @@
-import yaml
 import logging
 import asyncio
 import typing
 import collections
 
+import yaml
 import aiohttp
 from sqlalchemy.orm import Session
 from sqlalchemy.future import select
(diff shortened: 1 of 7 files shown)


rpisarev (commented):
lgtm

PR #80, opened 2022-01-19
ALBS-110: Fix empty streams based on mock options

Korulag:
This commit adds ability to set empty streams for build time and runtime
modules dependencies in modules.yaml file to correct ones from mock
options.

### DIFF
--- a/alws/build_planner.py
+++ b/alws/build_planner.py
@@ -154,7 +154,12 @@ class BuildPlanner:
             settings.pulp_password
         )
         module = None
-        mock_options = {'definitions': {}}
+        if self._build.mock_options:
+            mock_options = self._build.mock_options.copy()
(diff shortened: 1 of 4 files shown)


rpisarev (approved):

alws/build_planner.py
```diff
@@ -154,7 +154,12 @@ async def add_task(self, task: build_schema.BuildTaskRef):
             settings.pulp_password
         )
         module = None
-        mock_options = {'definitions': {}}
+        if self._build.mock_options:
+            mock_options = self._build.mock_options.copy()
+            if not mock_options.get('definitions'):
```
soksanichenko, line 159: is not None

anfimovdm (approved):

PR #81, opened 2022-01-20
Albs 73 with repodata exporting

rpisarev:
Added a couple of libs. Also changed some volumes (web-server have to store repodata near exported repos by pulp). So the docker images albs-web-server_web_server_1 should be rebuilt.

For pulp config pulp/setting/setting.py we should use ALLOWED_EXPORT_PATHS=['/srv/exports'] line to allow exporting

### DIFF
--- a/alws/routers/repositories.py
+++ b/alws/routers/repositories.py
@@ -49,11 +49,7 @@ async def fs_export_repository(repository_ids: list,
         res = await get_urls_from_html(repo_url)
         for url in res:
             dir_rd = Path(repo_elem).parent / 'repodata'
-            print(dir_rd)
-            try:
-                os.makedirs(dir_rd, exist_ok=True)
-            except Exception as e:
-                pass
+            os.makedirs(dir_rd, exist_ok=True)
             await download_file(url, dir_rd / Path(url).name)
     return export_paths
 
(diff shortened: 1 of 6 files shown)


anfimovdm (approved):

alws/routers/repositories.py
```diff
@@ -35,6 +41,33 @@ async def fs_export_repository(repository_ids: list,
                                db: database.Session = Depends(get_db)):
     export_task = await repo_exporter.create_pulp_exporters_to_fs(
         db, repository_ids)
-    export_paths = await repo_exporter.execute_pulp_exporters_to_fs(
+    export_data = await repo_exporter.execute_pulp_exporters_to_fs(
         db, export_task)
+    export_paths = list(export_data.keys())
```
monkrobot, line 46: Maybe it's better to move this and the functions below to another file? It's not what you wait to see in routers

alws/utils/pulp_client.py
```diff
@@ -327,7 +328,7 @@ async def create_filesystem_exporter(self, fse_name: str, fse_path: str,
     async def update_filesystem_exporter(self, fse_pulp_href: str,
                                          fse_name: str,
                                          fse_path: str,
-                                         fse_method: str='symlink'):
+                                         fse_method: str='hardlink'):
```
monkrobot, line 331: what type is returned?

monkrobot (approved):

soksanichenko (approved):

PR #100, opened 2022-02-02
ALBS-118. Added data migration to restore signkey for build artifacts.

rpisarev:
Tested upgrading and downgraded too.
Based on ALBS-115 branch


alws/alembic/versions/6f8e782397d2_data_migration_added_sign_keys_for_.py
```diff
@@ -0,0 +1,78 @@
...
+                        sign_key_id=sign_key.id
+                )
+            )
+        db.commit()
+    print('upgraded!')
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    print('downgrading...')
```
anfimovdm, line 77: `downgrade()` func seems like useless, is that right behavior?
I'm think would be better if you make script in /scripts/ than `alembic` migration, because we doesn't change DB structure right here

alws/alembic/versions/6f8e782397d2_data_migration_added_sign_keys_for_.py
```diff
@@ -0,0 +1,78 @@
...
+                models.BuildTask,
+                models.BuildTask.id == models.BuildTaskArtifact.build_task_id
+            ).where(
+                models.BuildTask.build_id == sign_task.build.id
+            ).filter(and_(
+                models.BuildTaskArtifact.type == 'rpm',
+                models.BuildTaskArtifact.sign_key == None
+            ))
+            res = db.execute(build_task_q)
+            for _, artifact_id, _, _ in res:
```
anfimovdm, line 60: can we select only use `models.BuildTaskArtifact.id` and iterate by them?

alws/alembic/versions/6f8e782397d2_data_migration_added_sign_keys_for_.py
```diff
@@ -0,0 +1,78 @@
...
+                models.BuildTaskArtifact.sign_key_id,
+                models.BuildTask.build_id
+            ).join(
+                models.BuildTask,
+                models.BuildTask.id == models.BuildTaskArtifact.build_task_id
+            ).where(
+                models.BuildTask.build_id == sign_task.build.id
+            ).filter(and_(
+                models.BuildTaskArtifact.type == 'rpm',
+                models.BuildTaskArtifact.sign_key == None
```
anfimovdm, line 57: can we union `where` and `filter` in one `filter/where` method?

anfimovdm (changes requested):

alws/alembic/versions/6f8e782397d2_data_migration_added_sign_keys_for_.py
rpisarev: @Korulag asked me create it exactly via migration instead of script I wanted too

alws/alembic/versions/6f8e782397d2_data_migration_added_sign_keys_for_.py
rpisarev: I will check it, thank you!

alws/alembic/versions/6f8e782397d2_data_migration_added_sign_keys_for_.py
rpisarev: It's a good question! I will check it too, thanks!

monkrobot (approved):

soksanichenko (approved):

PR #108, opened 2022-02-10
ALBS-144: Added logic in releaser for correct placing in pulp repos

anfimovdm:
- Added table `reference_platforms`
- Changed script for loading reference platforms
- Added retry for multilib
- Fixed schemas for reference platforms
- Fixed create/modify reference platforms
- Fixed schema for sign key create

### NOTE:
After applying alembic migration, need to update platforms in DB for filling field `is_reference`


alws/models.py
```diff
@@ -101,12 +101,31 @@ class Platform(Base):
         unique=True,
         index=True
     )
+    reference_platform = sqlalchemy.Column(
+        sqlalchemy.VARCHAR(length=20), nullable=True)
```
AvoidMe, line 105: It seems to me, what it can be more than one reference_platform, so it's better to be many-to-many relationship

alws/models.py
```diff
@@ -101,12 +101,31 @@ class Platform(Base):
...
+    type = sqlalchemy.Column(sqlalchemy.Text, nullable=False)
+    distr_type = sqlalchemy.Column(sqlalchemy.Text, nullable=False)
+    distr_version = sqlalchemy.Column(sqlalchemy.Text, nullable=False)
+    arch_list = sqlalchemy.Column(JSONB, nullable=False)
+    name = sqlalchemy.Column(
+        sqlalchemy.Text,
+        nullable=False,
+        unique=True,
+        index=True,
+    )
```
AvoidMe, line 126: It seems to me, that this data is mostly duplicate to `Platform`, can we use just platform with many-to-many relationships to itself instead?

AvoidMe (changes requested):

alws/models.py
anfimovdm: Thanks, fixed

alws/models.py
anfimovdm: Fixed

PR #118, opened 2022-02-21
ALBS-152: Replace async calls to DB by sync

soksanichenko:
- FastAPI dependency `get_db` is replaced by contextmanager `get_sync_db` in all entries
- Async connection to db is replaced by sync connection.


anfimovdm (approved):
There is also some scripts in `scripts/` that uses async sessions, but i'm think we can change them later

Korulag:
Looks like we are adapted to using async API without causing strange issues, so this PR is not needed now

PR #132, opened 2022-03-10
Fixed modules release

Korulag:


alws/crud/release.py
```diff
@@ -127,6 +128,19 @@ def get_pulp_based_response():
             'modules': rpm_modules,
         }
 
+    if not settings.package_beholder_enabled:
+        return get_pulp_based_response()
+
+    clean_base_dist_name = re.search(
+        r'(?P<dist_name>[a-z]+)', base_dist_name,
+        re.IGNORECASE).groupdict().get('dist_name')
+    clean_base_dist_name_lower = clean_base_dist_name.lower()
```
soksanichenko, line 137: You'll get exception if `clean_base_dist_name` is None

alws/crud/release.py
```diff
@@ -127,6 +128,19 @@ def get_pulp_based_response():
...
+    clean_ref_dist_name = re.search(
+        r'(?P<dist_name>[a-z]+)', reference_dist_name,
+        re.IGNORECASE).groupdict().get('dist_name')
+    clean_ref_dist_name_lower = clean_ref_dist_name.lower()
```
soksanichenko, line 141: same as above

alws/crud/release.py
Korulag: Should not be, but you're right

anfimovdm (approved):

alws/crud/release.py
```diff
@@ -147,33 +167,40 @@ def get_pulp_based_response():
...
+        except Exception:
+            endpoint = (
+                f'/api/v1/distros/{base_dist_name}/'
+                f'{reference_dist_version}/module/{module["name"]}/'
+                f'{module["stream"]}/{module["arch"]}/'
+            )
+            try:
+                module_response = await beholder.get(endpoint)
+            except Exception:
+                success = False
```
AvoidMe, line 190: can be for loop with instead of manually handling exception two times
smth like this:
```
success = False
for endpoint in [A, B]:
   try:
     module_response = await beholder.get(endpoint)
  except Exception:
    continue
  success = True
```

PR #134, opened 2022-03-11
ALBS-188 Build with incorrect arch by using API

monkrobot:


alws/schemas/build_schema.py
```diff
@@ -72,7 +72,8 @@ class BuildTaskModuleRef(BaseModel):
 class BuildCreatePlatforms(BaseModel):
 
     name: str
-    arch_list: typing.List[str]
+    arch_list: typing.List[typing.Literal['x86_64', 'i868', 'aarch64', 'ppc64le',
```
anfimovdm, line 75: typo in `i868`?

anfimovdm (changes requested):

alws/schemas/build_schema.py
Korulag: definitely a typo, it's i686

monkrobot: Thx! Fix

anfimovdm (approved):

PR #146, opened 2022-03-22
ALBS-193: Added ability to updating modules and comps in pulp repos

anfimovdm:


alws/utils/file_utils.py
```diff
@@ -22,3 +23,9 @@ async def download_file(url: str, dest: BinaryIO):
         async with session.get(url) as response:
             content = await response.content.read()
             dest.write(content)
+
+
+def hash_content(content):
+    hasher = hashlib.new('sha256')
+    hasher.update(content.encode())
```
Korulag, line 30: You sure `encode` is needed here? AFAIK text could be passed straight into hasher

Korulag: However, I see that it's just copy of existing function into different place, so can skip it

Korulag (approved):

monkrobot (approved):

PR #149, opened 2022-03-25
Fixed build_done method. Improved reliability of task finish

Korulag:
This commit brings several things to make `build_done` endpoint more reliable:
1. Separate exceptions for different things: help react on a particular error;
2. Increased timeout for the `build_done` dramatiq actor;
3. Reworked build artifacts processing: now packages are added to the repository only when all of them were converted from artifacts to actual RPM packages;
4. Improved logs handling: logs are processed similarly to packages, but now they are first in line and committed separately from packages;
5. Modules errors handling: both parsing initial data and adding new artifacts.

### DIFF
--- a/alws/constants.py
+++ b/alws/constants.py
@@ -6,10 +6,11 @@ from collections import namedtuple
 
 __all__ = ['BuildTaskStatus', 'ReleaseStatus', 'TestTaskStatus',
            'BuildTaskRefType', 'SignStatus', 'RepoType', 'ExportStatus',
-           'debuginfo_regex', 'REQUEST_TIMEOUT']
+           'debuginfo_regex', 'REQUEST_TIMEOUT', 'DRAMATIQ_TASK_TIMEOUT']
 
 
 REQUEST_TIMEOUT = 60  # 1 minute
+DRAMATIQ_TASK_TIMEOUT = 3600000  # 1 hour in milliseconds
 
 
 class BuildTaskStatus(enum.IntEnum):
--- a/alws/crud/build_node.py
+++ b/alws/crud/build_node.py
@@ -219,7 +219,6 @@ async def __process_logs(pulp_client: PulpClient, task_id: int,
     hrefs = [item[0] for item in results]
     try:
         await pulp_client.modify_repository(repo.pulp_href, add=hrefs)
-        await pulp_client.create_file_publication(repo.pulp_href)
     except Exception as e:
         logging.error('Cannot add log files to the repository: %s',
                       str(e))
--- a/alws/crud/build_node.py
+++ b/alws/crud/build_node.py
@@ -10,6 +10,14 @@ from sqlalchemy.orm import Session, selectinload
 from alws import models
 from alws.config import settings
 from alws.constants import BuildTaskStatus
+from alws.errors import (
+    ArtifactConversionError,
+    ModuleUpdateError,
+    MultilibProcessingError,
+    NoarchProcessingError,
+    RepositoryAddError,
+    SrpmProvisionError,
+)
 from alws.schemas import build_node_schema
 from alws.utils.modularity import IndexWrapper
 from alws.utils.multilib import add_multilib_packages, get_multilib_packages
@@ -101,6 +109,8 @@ async def check_build_task_is_finished(db: Session, task_id: int) -> bool:
 async def __process_rpms(pulp_client: PulpClient, task_id: int, task_arch: str,
                          task_artifacts: list, repositories: list,
                          built_srpm_url: str = None, module_index=None):
+    repo_packages_mapping = {}
+    exceptions = []
     rpms = []
     for artifact in task_artifacts:
         arch = task_arch
@@ -116,14 +126,14 @@ async def __process_rpms(pulp_client: PulpClient, task_id: int, task_arch: str,
         )
         try:
             href = await pulp_client.create_rpm_package(
-                artifact.name, artifact.href, repo.pulp_href)
-            if module_index:
-                for module in module_index.iter_modules():
-                    rpm_package = await pulp_client.get_rpm_package(href)
-                    module.add_rpm_artifact(rpm_package)
+                artifact.name, artifact.href)
+            if repo.pulp_href not in repo_packages_mapping:
+                repo_packages_mapping[repo.pulp_href] = []
+            repo_packages_mapping[repo.pulp_href].append(href)
         except Exception as e:
             logging.error('Cannot create RPM package for %s, error: %s',
                           str(artifact), str(e))
+            exceptions.append(e)
             continue
         else:
             rpms.append(
@@ -134,11 +144,48 @@ async def __process_rpms(pulp_client: PulpClient, task_id: int, task_arch: str,
                     href=href
                 )
             )
+    if exceptions:
+        raise ArtifactConversionError(
+            'Cannot put RPM packages into Pulp storage: %s',
+            '\n'.join([str(e) for e in exceptions]))
+    for repo_href, packages in repo_packages_mapping.items():
+        try:
+            await pulp_client.modify_repository(repo_href, add=packages)
+        except Exception as e:
+            logging.error('Cannot add RPM packages to the repository: %s',
+                          str(e))
+            exceptions.append(e)
+        try:
+            await pulp_client.create_rpm_publication(repo_href)
+        except Exception as e:
+            logging.error('Cannot create a publication for the repository: %s',
+                          str(e))
+            exceptions.append(e)
+    if exceptions:
+        raise RepositoryAddError(
+            'Cannot save RPM packages into Pulp repositories: %s',
+            '\n'.join([str(e) for e in exceptions]))
+
+    if module_index:
+        packages_info = {}
+        for rpm in rpms:
+            rpm_package = await pulp_client.get_rpm_package(rpm.href)
+            packages_info[rpm.id] = rpm_package
+        try:
+            for module in module_index.iter_modules():
+                for rpm in rpms:
+                    rpm_package = packages_info[rpm.id]
+                    module.add_rpm_artifact(rpm_package)
+        except Exception as e:
+            raise ModuleUpdateError('Cannot update module: %s', str(e)) from e
+
     return rpms
 
 
 async def __process_logs(pulp_client: PulpClient, task_id: int,
                          task_artifacts: list, repositories: list):
+    repo_files_mapping = {}
+    exceptions = []
     logs = []
     str_task_id = str(task_id)
     for artifact in task_artifacts:
@@ -148,10 +195,14 @@ async def __process_logs(pulp_client: PulpClient, task_id: int,
         )
         try:
             href = await pulp_client.create_file(
-                artifact.name, artifact.href, repo.pulp_href)
+                artifact.name, artifact.href)
+            if repo.pulp_href not in repo_files_mapping:
+                repo_files_mapping[repo.pulp_href] = []
+            repo_files_mapping[repo.pulp_href].append(href)
         except Exception as e:
             logging.error('Cannot create log file for %s, error: %s',
                           str(artifact), str(e))
+            exceptions.append(e)
             continue
         else:
             logs.append(
@@ -162,6 +213,29 @@ async def __process_logs(pulp_client: PulpClient, task_id: int,
                     href=href
                 )
             )
+    if exceptions:
+        raise ArtifactConversionError(
+            'Cannot put build logs into Pulp storage: %s',
+            '\n'.join([str(e) for e in exceptions]))
+
+    for repo_href, files in repo_files_mapping.items():
+        try:
+            await pulp_client.modify_repository(repo_href, add=files)
+        except Exception as e:
+            logging.error('Cannot add log files to the repository: %s',
+                          str(e))
+            exceptions.append(e)
+            continue
+        try:
+            await pulp_client.create_file_publication(repo_href)
+        except Exception as e:
+            logging.error('Cannot create publication for log files: %s',
+                          str(e))
+            exceptions.append(e)
+    if exceptions:
+        raise RepositoryAddError(
+            'Cannot save build log into Pulp repository: %s',
+            '\n'.join([str(e) for e in exceptions]))
     return logs
 
 
@@ -198,7 +272,9 @@ async def __process_build_task_artifacts(
                 module_repo.url)
             module_index = IndexWrapper.from_template(repo_modules_yaml)
         except Exception as e:
+            message = f'Cannot parse modules index: {str(e)}'
             logging.error('Cannot parse modules index: %s', str(e))
+            raise ModuleUpdateError(message) from e
     rpm_artifacts = [item for item in task_artifacts if item.type == 'rpm']
     log_artifacts = [item for item in task_artifacts
                      if item.type == 'build_log']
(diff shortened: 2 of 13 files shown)


alws/crud/build_node.py
```diff
@@ -134,11 +144,48 @@ async def __process_rpms(pulp_client: PulpClient, task_id: int, task_arch: str,
...
+            '\n'.join([str(e) for e in exceptions]))
+    for repo_href, packages in repo_packages_mapping.items():
+        try:
+            await pulp_client.modify_repository(repo_href, add=packages)
+        except Exception as e:
+            logging.error('Cannot add RPM packages to the repository: %s',
+                          str(e))
+            exceptions.append(e)
+        try:
+            await pulp_client.create_rpm_publication(repo_href)
```
anfimovdm, line 159: Build rpm repos triggers rpm publication after adding/removing content, so may be we don't need this?

alws/dramatiq/build.py
```diff
@@ -63,7 +71,11 @@ def start_build(build_id: int, build_request: Dict[str, Any]):
 
 @dramatiq.actor(
     max_retries=0,
-    priority=1
+    priority=1,
+    time_limit=3600000,
```
anfimovdm, line 75: Pls, add comment for this value from your latest changes for dramatiq

anfimovdm (changes requested):

alws/crud/build_node.py
```diff
@@ -162,6 +213,29 @@ async def __process_logs(pulp_client: PulpClient, task_id: int,
...
+            logging.error('Cannot add log files to the repository: %s',
+                          str(e))
+            exceptions.append(e)
+            continue
+        try:
+            await pulp_client.create_file_publication(repo_href)
+        except Exception as e:
+            logging.error('Cannot create publication for log files: %s',
+                          str(e))
+            exceptions.append(e)
```
anfimovdm, line 234: Adding logs in repo, seems like logic for adding rpms/modules in repos, can we use it like seperate func?

alws/crud/build_node.py
Korulag: That's exactly what I've got rid of. Slava pointed out that such approach triggers 3 tasks per each package in pulp (3*N) while modernized approach is N + 2, where N is number of packages and 2 - adding content to repository + making publication. In order to achieve it we first do just artifacts -> RPMs conversion, then add RPMs to repository, then manually trigger publication

alws/dramatiq/build.py
Korulag: Yeah, probably need to move it to constants

alws/crud/build_node.py
anfimovdm: Okay, so may be we need to turn off autopublish for rpm repos?

alws/crud/build_node.py
Korulag: Yes, no, maybe... They use different error messages, but I'll try to move the logic into separate function

anfimovdm: It's up to you anyway)

alws/crud/build_node.py
AvoidMe: Agreed with @anfimovdm, why we need to remove autopublish from repos? It's actually +1 request for every repository, which makes whole thing slower

alws/crud/build_node.py
```diff
@@ -134,11 +144,48 @@ async def __process_rpms(pulp_client: PulpClient, task_id: int, task_arch: str,
                     href=href
                 )
             )
+    if exceptions:
+        raise ArtifactConversionError(
+            'Cannot put RPM packages into Pulp storage: %s',
+            '\n'.join([str(e) for e in exceptions]))
+    for repo_href, packages in repo_packages_mapping.items():
+        try:
+            await pulp_client.modify_repository(repo_href, add=packages)
```
AvoidMe, line 153: We can make this requests in parallel, so whole thing will be much faster on large number of repositories

alws/crud/build_node.py
```diff
@@ -134,11 +144,48 @@ async def __process_rpms(pulp_client: PulpClient, task_id: int, task_arch: str,
...
+    if exceptions:
+        raise RepositoryAddError(
+            'Cannot save RPM packages into Pulp repositories: %s',
+            '\n'.join([str(e) for e in exceptions]))
+
+    if module_index:
+        packages_info = {}
+        for rpm in rpms:
+            rpm_package = await pulp_client.get_rpm_package(rpm.href)
+            packages_info[rpm.id] = rpm_package
```
AvoidMe, line 173: Same as above, this can be moved to `asyncio.gather()`, which will seriously improve overall execution speed
Also, maybe we should consider to make single request for all packages using `pkgId__in` (which is actually sha256)
https://docs.pulpproject.org/pulp_rpm/restapi.html#operation/content_rpm_packages_list

alws/crud/build_node.py
```diff
@@ -162,6 +213,29 @@ async def __process_logs(pulp_client: PulpClient, task_id: int,
...
                 )
             )
+    if exceptions:
+        raise ArtifactConversionError(
+            'Cannot put build logs into Pulp storage: %s',
+            '\n'.join([str(e) for e in exceptions]))
+
+    for repo_href, files in repo_files_mapping.items():
+        try:
+            await pulp_client.modify_repository(repo_href, add=files)
```
AvoidMe, line 223: Same as above, can be made in parallel, publication can be removed

AvoidMe (changes requested):

alws/crud/build_node.py
Korulag: As for autopublish: it creates several tasks per each package while this approach will only make 1 call for new repo version/publication after all packages added

Korulag: Also, each task has only a handful of repositories, basically 3: src, arch, debug. Not sure if parallelization here will make any difference

alws/crud/build_node.py
Korulag: Yeah, found the thing about publication, removing that call

alws/crud/build_node.py
Korulag: We do not have SHA256 info in this part and adding it will need modifications on build node side

alws/utils/pulp_client.py
```diff
@@ -466,3 +501,13 @@ async def request(
                 response_json = await response.json()
                 response.raise_for_status()
                 return response_json
+
+
+async def create_entity(pulp_client: PulpClient, artifact):
```
anfimovdm, line 506: I'm think we should turn this func into method of `PulpClient` class

alws/crud/build_node.py
```diff
@@ -98,62 +107,107 @@ async def check_build_task_is_finished(db: Session, task_id: int) -> bool:
...
+    processed_packages = []
+    for tasks, repo in (
+            (src_packages_tasks, src_repo), (arch_packages_tasks, arch_repo),
+            (debug_packages_tasks, debug_repo)):
+        if tasks:
+            try:
+                results = await asyncio.gather(*tasks)
+            except Exception as e:
+                logging.exception('Cannot create RPM packages for error: %s',
+                                  str(e))
```
anfimovdm, line 152: There is no need to using args for `logger.exception`

alws/crud/build_node.py
```diff
@@ -98,62 +107,107 @@ async def check_build_task_is_finished(db: Session, task_id: int) -> bool:
...
+                raise ArtifactConversionError(
+                    'Cannot put RPM packages into Pulp storage: %s', str(e))
+            else:
+                processed_packages.extend(results)
+                hrefs = [item[0] for item in results]
+                try:
+                    await pulp_client.modify_repository(
+                        repo.pulp_href, add=hrefs)
+                except Exception as e:
+                    logging.error('Cannot add RPM packages to the repository: %s',
```
anfimovdm, line 162: Pls, use `logger.exception`

alws/crud/build_node.py
```diff
@@ -98,62 +107,107 @@ async def check_build_task_is_finished(db: Session, task_id: int) -> bool:
...
+    repo = next(
+        repo for repo in repositories
+        if repo.name.endswith(str_task_id)
+    )
+    files = [create_entity(pulp_client, artifact)
+             for artifact in task_artifacts]
+    try:
+        results = await asyncio.gather(*files)
+    except Exception as e:
+        logging.error('Cannot create log files for %s, error: %s',
```
anfimovdm, line 205: `logging.exception` instead of `logging.error`

alws/crud/build_node.py
```diff
@@ -98,62 +107,107 @@ async def check_build_task_is_finished(db: Session, task_id: int) -> bool:
...
-                artifact.name, artifact.href, repo.pulp_href)
-            if module_index:
-                for module in module_index.iter_modules():
-                    rpm_package = await pulp_client.get_rpm_package(href)
+            for module in module_index.iter_modules():
+                for rpm in rpms:
+                    rpm_package = packages_info[rpm.href]
                     module.add_rpm_artifact(rpm_package)
         except Exception as e:
-            logging.error('Cannot create RPM package for %s, error: %s',
```
anfimovdm, line 125: Why we don't use logging for module exceptions?

alws/crud/build_node.py
```diff
@@ -162,6 +216,14 @@ async def __process_logs(pulp_client: PulpClient, task_id: int,
                     href=href
                 )
             )
+    hrefs = [item[0] for item in results]
```
anfimovdm, line 219: There is no need to use list comprehension for retrieving hrefs, above you iterate on results of async tasks and you can append hrefs in some list

alws/crud/distribution.py
```diff
@@ -104,41 +106,66 @@ async def add_distributions_after_rebuild(
             await pulp_client.create_rpm_publication(repo)
 
 
+async def get_existing_packages(pulp_client: PulpClient,
+                                repository: models.Repository):
+    return await pulp_client.get_rpm_repository_packages(
+        repository.pulp_href,
+        include_fields=['pulp_href', 'artifact', 'sha256', 'location_href',
+                        'arch'])
```
anfimovdm, line 114: I'm think this func can be usefull for some tasks in future, can we turn it into method for `PulpClient` class?
This func looks kinda strange, you can make it look better, but it's up to you

alws/crud/distribution.py
```diff
@@ -152,38 +179,44 @@ async def modify_distribution(build_id: int, distribution: str, db: Session,
...
+    if modification == 'add':
+        db_distro.builds.append(db_build)
+    else:
+        remove_query = models.Build.id.__eq__(build_id)
+        await db.execute(
+            delete(models.DistributionBuilds).where(remove_query)
+        )
+    db.add(db_distro)
+    try:
+        await db.commit()
```
anfimovdm, line 220: If there will be exception during `db.commit`, does session rollback working? or it's should be called manually?

alws/crud/sign_task.py
```diff
@@ -221,6 +220,7 @@ async def complete_sign_task(db: Session, sign_task_id: int,
 
             for repo_href, packages in packages_to_add.items():
                 await pulp_client.modify_repository(repo_href, add=packages)
```
anfimovdm, line 222: I'm think we can modify/publish it in `asyncio.gather`

alws/utils/modularity.py
```diff
@@ -15,10 +13,6 @@
 from gi.repository import Modulemd
 
 
-def get_random_unique_version():
```
anfimovdm, line 18: I may be wrong, but this func using also for repodata uploader and somewhere else, pls check it

anfimovdm (changes requested):

alws/crud/build_node.py
```diff
@@ -98,62 +107,107 @@ async def check_build_task_is_finished(db: Session, task_id: int) -> bool:
...
-            build_repo for build_repo in repositories
-            if build_repo.arch == arch
-            and build_repo.type == artifact.type
-            and build_repo.debug == artifact.is_debuginfo
+        if artifact.arch == 'src' and built_srpm_url is None:
+            src_packages_tasks.append(create_entity(pulp_client, artifact))
+        if artifact.is_debuginfo:
+            debug_packages_tasks.append(create_entity(pulp_client, artifact))
+        else:
+            arch_packages_tasks.append(create_entity(pulp_client, artifact))
```
AvoidMe, line 141: Src packages will be added twice, right?

alws/crud/build_node.py
```diff
@@ -98,62 +107,107 @@ async def check_build_task_is_finished(db: Session, task_id: int) -> bool:
...
+
+    processed_packages = []
+    for tasks, repo in (
+            (src_packages_tasks, src_repo), (arch_packages_tasks, arch_repo),
+            (debug_packages_tasks, debug_repo)):
+        if tasks:
+            try:
+                results = await asyncio.gather(*tasks)
```
AvoidMe, line 149: why not all of them together?

alws/constants.py
```diff
@@ -6,10 +6,11 @@
 
 __all__ = ['BuildTaskStatus', 'ReleaseStatus', 'TestTaskStatus',
            'BuildTaskRefType', 'SignStatus', 'RepoType', 'ExportStatus',
-           'debuginfo_regex', 'REQUEST_TIMEOUT']
+           'debuginfo_regex', 'REQUEST_TIMEOUT', 'DRAMATIQ_TASK_TIMEOUT']
 
 
 REQUEST_TIMEOUT = 60  # 1 minute
+DRAMATIQ_TASK_TIMEOUT = 3600000  # 1 hour in milliseconds
```
soksanichenko, line 13: Just for future: I guess better to write multiplication of some constants like this: 60*60*1000
It's more readable

alws/crud/build_node.py
soksanichenko, line 152: Conversation to str is not needed. It will be done automatically.

alws/crud/distribution.py
```diff
@@ -104,41 +106,66 @@ async def add_distributions_after_rebuild(
...
+        repository.pulp_href,
+        include_fields=['pulp_href', 'artifact', 'sha256', 'location_href',
+                        'arch'])
+
+
+async def get_packages_to_add(
+        pulp_client: PulpClient, build_repo: models.Repository,
+        dist_repo: models.Repository):
+
+    def filter_by_arch(pkgs: typing.List[dict], repo_arch: str):
```
soksanichenko, line 121: I saw the same func somewhere in the project. May be better to move it to some util.py module?)

alws/crud/distribution.py
```diff
@@ -104,41 +106,66 @@ async def add_distributions_after_rebuild(
...
+            if pkg['arch'] == 'noarch' and repo_arch != 'src':
+                filtered.append(pkg)
+            elif pkg['arch'] == 'i686' and repo_arch in ('i686', 'x86_64'):
+                filtered.append(pkg)
+            elif pkg['arch'] == repo_arch:
+                filtered.append(pkg)
+        return filtered
+
+    dist_packages = await get_existing_packages(pulp_client, dist_repo)
+    search_by_href = set([p['pulp_href'] for p in dist_packages])
```
soksanichenko, line 133: List is not needed

alws/crud/distribution.py
soksanichenko: Ideally, we should do it in the separate context manager)

soksanichenko (changes requested):

alws/crud/distribution.py
Korulag: Ideally...

alws/crud/build_node.py
Korulag: Each set of artifacts needs to be converted and then added to the corresponding repo. Tried to not overcomplicate logic here

alws/crud/build_node.py
Korulag: You're right, need to fix the condition

soksanichenko (approved):

PR #150, opened 2022-03-28
ALBS-224: Improvements for releases

anfimovdm:
- Transfer release logic in class
- Added checking packages presence in repos for all packages and on every stage of release


alws/release_planner.py
```diff
@@ -0,0 +1,704 @@
...
+            ),
+            selectinload(models.Build.repos)
+        )
+        build_result = await self._db.execute(builds_q)
+        modules_to_release = {}
+        for build in build_result.scalars().all():
+            for rpm_data in (build.source_rpms, build.binary_rpms):
+                for rpm in rpm_data:
+                    # Failsafe to not process logs
+                    if rpm.artifact.type != 'rpm':
```
Korulag, line 74: AFAIK neither binary_rpm nor source_rpm can contain logs

alws/release_planner.py
```diff
@@ -0,0 +1,704 @@
...
+        arch: str,
+    ):
+        params = {
+            'name__in': ','.join(pkgs_nevra['name']),
+            'epoch__in': ','.join(pkgs_nevra['epoch']),
+            'version__in': ','.join(pkgs_nevra['version']),
+            'release__in': ','.join(pkgs_nevra['release']),
+            'arch': arch,
+            'repository_version': repo_ver_href,
+            'fields': 'pulp_href,name,epoch,version,release,arch',
```
Korulag, line 138: fields could be passed as list

Korulag (changes requested):

alws/release_planner.py
anfimovdm: recently i've found some interesting thing, if we pass in pulp list of str for some params, something going wrong in pulp and query looks like this:
`rpm/packages/?fields=pulp_href&?fields=name&?fields=epoch&?fields=version...`
it's look like a bug or something, and i'm afraid that with large params of lists, in very bad case we can exceed limit of query path

alws/release_planner.py
```diff
@@ -0,0 +1,704 @@
...
+                models.BinaryRpm.artifact),
+            selectinload(models.Build.tasks).selectinload(
+                models.BuildTask.rpm_module
+            ),
+            selectinload(models.Build.repos)
+        )
+        build_result = await self._db.execute(builds_q)
+        modules_to_release = {}
+        for build in build_result.scalars().all():
+            for rpm_data in (build.source_rpms, build.binary_rpms):
```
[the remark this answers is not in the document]
anfimovdm: Yeah, you're right, thanks

alws/release_planner.py
anfimovdm: It's old logic from releases, i'll check it

alws/alembic/versions/077264687344_added_priority_arches_to_copy_in_release.py
```diff
@@ -0,0 +1,28 @@
...
+Create Date: 2022-03-29 10:01:32.470281
+
+"""
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision = '077264687344'
+down_revision = '5c3da459c12f'
```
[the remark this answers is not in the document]
anfimovdm: it's already rebased

Korulag (approved):

PR #151, opened 2022-03-29
Fixed build_done #2. Fixed logs handling and multilib

Korulag:


alws/crud/build_node.py
```diff
@@ -159,8 +159,8 @@ def get_repo(repo_arch, is_debug):
                     await pulp_client.modify_repository(
                         repo.pulp_href, add=hrefs)
                 except Exception as e:
-                    logging.error('Cannot add RPM packages to the repository: %s',
-                                  str(e))
+                    logging.exception('Cannot add RPM packages '
+                                      'to the repository: %s', e)
```
anfimovdm, line 163: there should be repo name instead of exception

alws/crud/build_node.py
```diff
@@ -175,8 +175,8 @@ def get_repo(repo_arch, is_debug):
         )
 
     if module_index:
-        results = await asyncio.gather(*[__get_rpm_package_info(
-            pulp_client, rpm.href) for rpm in rpms])
+        results = await asyncio.gather((__get_rpm_package_info(
+            pulp_client, rpm.href) for rpm in rpms))
```
anfimovdm, line 179: not sure what `gather` can unpack generator

alws/utils/pulp_client.py
```diff
@@ -513,25 +523,16 @@ async def request(
...
-
-async def create_entity(pulp_client: PulpClient, artifact):
-    if artifact.type == 'rpm':
-        entity_href = await pulp_client.create_rpm_package(
-            artifact.name, artifact.href)
-    else:
-        entity_href = await pulp_client.create_file(
-            artifact.name, artifact.href)
-    return entity_href, artifact
+            async with RetryClient() as client:
```
anfimovdm, line 526: i'm think we should define statuses to retry

anfimovdm (changes requested):

alws/crud/build_node.py
Korulag: It actually unpacks it, tested locally

PR #152, opened 2022-03-30
Added platform for AlmaLinux 8 distribution

Korulag:


Korulag:
Implemented this functionality with flavors

PR #170, opened 2022-04-13
ALBS-258 Add pageNumber query parameter in GET releases endpoint

monkrobot:


alws/crud/release.py
```diff
@@ -2,13 +2,25 @@
...
         selectinload(models.Release.platform),
-    ).order_by(models.Release.id))
+    ).order_by(models.Release.id)
+    if page_number:
+        query = query.slice(10 * page_number - 10, 10 * page_number)
+    release_result = await db.execute(query)
+    if page_number:
+        total_releases = await db.execute(func.count(models.Release.id))
+        total_releases = total_releases.scalar()
+        return {'builds': release_result.scalars().all(),
```
anfimovdm, line 23: it's should be `releases` instead of `builds`

alws/schemas/release_schema.py
```diff
@@ -28,6 +28,13 @@ class Config:
         orm_mode = True
 
 
+class ReleaseResponse(BaseModel):
+
+    builds: typing.List[Release]
```
anfimovdm, line 33: same as above

anfimovdm (changes requested):

alws/crud/release.py
monkrobot: Thanks! Fix

alws/schemas/release_schema.py
monkrobot: Fix

anfimovdm (approved):

alws/crud/release.py
```diff
@@ -2,13 +2,25 @@
...
-async def get_releases(db: Session) -> typing.List[models.Release]:
-    release_result = await db.execute(select(models.Release).options(
+async def get_releases(page_number: typing.Optional[int],
+                       db: Session
+) -> typing.List[models.Release]:
+    query = select(models.Release).options(
         selectinload(models.Release.created_by),
         selectinload(models.Release.platform),
-    ).order_by(models.Release.id))
+    ).order_by(models.Release.id)
```
maccelf, line 16: Add `models.Release.id.desc()` instead `models.Release.id` so that releases are in the correct order

maccelf (changes requested):

alws/crud/release.py
monkrobot: Done

maccelf (approved):

PR #176, opened 2022-04-19
Added separate setting to set Pulp internal host address

Korulag:


Korulag:
There are issues with using this approach, so closing PR for now until we find a better way

PR #190, opened 2022-05-04
Verify packages signatures during export from Pulp

Korulag:


soksanichenko (approved):

anfimovdm (changes requested):

scripts/packages_exporter.py
```diff
@@ -144,6 +162,14 @@ async def get_sign_keys(self):
         endpoint = 'sign-keys/'
         return await self.make_request('GET', endpoint)
 
+    async def download_repodata(self, repodata_path, repodata_url):
+        file_links = await get_repodata_file_links(repodata_url)
+        for link in file_links:
+            file_name = os.path.basename(link)
+            if not link.endswith('/'):
```
soksanichenko, line 169: 1. You can always add slash to the end of url and it doesn't break anything
2. Use `from urllib.parse import urljoin` for joining uris. As I remember `os.path.join` can return incorrect result in some cases.

scripts/packages_exporter.py
```diff
@@ -156,6 +182,22 @@ async def export_repositories(self, repo_ids: list) -> typing.List[str]:
...
+            parent_dir = str(Path(export_path).parent)
+            repodata_path = os.path.abspath(os.path.join(
+                parent_dir, 'repodata'))
+            repodata_url = urllib.parse.urljoin(
+                exporter['repo_url'], 'repodata/')
+            try:
+                local['sudo']['chown', '-R',
+                              f'{self.current_user}:{self.current_user}',
+                              f'{parent_dir}'].run()
+                if not os.path.exists(repodata_path):
```
soksanichenko, line 194: For what do you check existing the directory if you use `exist_ok=True`?

scripts/packages_exporter.py
soksanichenko: Not actual

soksanichenko (approved):

anfimovdm (approved):

PR #194, opened 2022-05-12
Module artifacts fixes

Korulag:


alws/crud/build_node.py
```diff
@@ -350,6 +362,14 @@ async def __process_build_task_artifacts(
                         if repo.type == 'rpm']
     log_repositories = [repo for repo in repositories
                         if repo.type == 'build_log']
+    src_rpm = next((
+        artifact.name for artifact in task_artifacts
+        if artifact.arch == 'src' and artifact.type == 'rpm'
+    ), None)
+    if not src_rpm:
```
soksanichenko, line 369: src_rpm is not None

alws/crud/build_node.py
```diff
@@ -236,31 +238,41 @@ def get_repo(repo_arch, is_debug):
...
-            res = await db.execute(srpm_query)
-            srpms = res.scalars().all()
-            srpm_href = next(
-                i.artifact.href for i in srpms if i.artifact.name == srpm.name
-            )
-            srpm_info = await pulp_client.get_rpm_package(
-                srpm_href, include_fields=pkg_fields)
+            srpms = [item for item in task_artifacts if item.arch == 'src'
+                     and item.type == 'rpm']
+            if srpms:
```
soksanichenko, line 243: srpm = next(iter(srpms), None)

soksanichenko (approved):

alws/crud/build_node.py
```diff
@@ -58,7 +59,7 @@ async def get_available_build_task(
                         models.Build.platform_flavors).selectinload(
                         models.PlatformFlavour.repos),
                     selectinload(models.BuildTask.rpm_module)
-            ).order_by(models.BuildTask.id)
+            ).order_by(models.BuildTask.id.asc())
```
anfimovdm, line 62: It's necessary? because I'm think that if will affect us in future

alws/crud/build_node.py
anfimovdm: also we can wrap list comprehension/generator with `next` and don't use `iter`

alws/crud/build_node.py
```diff
@@ -236,31 +238,41 @@ def get_repo(repo_arch, is_debug):
...
+                    i.artifact.href for i in srpms
+                    if i.artifact.name == srpm.name
+                )
+                srpm_info = await pulp_client.get_rpm_package(
+                    srpm_href, include_fields=pkg_fields)
         try:
             for module in module_index.iter_modules():
                 for rpm in rpms:
+                    if rpm.name.endswith('.src.rpm'):
+                        modular_rpms.append(rpm)
```
anfimovdm, line 264: i can't understand why we need this

anfimovdm (changes requested):

alws/crud/build_node.py
Korulag: Should not affect, because this only ensures we will pick up the task with smallest ID

alws/crud/build_node.py
Korulag: We have some RPMs that should be filtered and not appear in repository at all. So we need to ensure that SRC rpms will be added anyway to the final list

anfimovdm (approved):

PR #199, opened 2022-05-19
Fixed partial modules update

Korulag:
This commit fixes partial modules update: do not add artifacts that are
not related to each requested architecture

### DIFF
--- a/alws/build_planner.py
+++ b/alws/build_planner.py
@@ -12,14 +12,19 @@ from alws.errors import DataNotFoundError, EmptyBuildError
 from alws.config import settings
 from alws.schemas import build_schema
 from alws.constants import BuildTaskStatus, BuildTaskRefType
-from alws.utils.pulp_client import PulpClient
-from alws.utils.parsing import parse_git_ref
-from alws.utils.modularity import (
-    ModuleWrapper, calc_dist_macro, IndexWrapper
-)
+from alws.utils.beholder_client import BeholderClient
 from alws.utils.gitea import (
     GiteaClient,
 )
+from alws.utils.modularity import (
+    calc_dist_macro,
+    IndexWrapper,
+    ModuleWrapper,
+    RpmArtifact,
+)
+from alws.utils.multilib import MultilibProcessor
+from alws.utils.parsing import parse_git_ref
+from alws.utils.pulp_client import PulpClient
 
 
 __all__ = ['BuildPlanner']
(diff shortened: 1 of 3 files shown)


alws/build_planner.py
```diff
@@ -135,16 +140,62 @@ async def init_build_repos(self):
...
+    ) -> typing.List[dict]:
+        beholder_client = BeholderClient(
+            settings.beholder_host, token=settings.beholder_token)
+        major_distr_version = task.module_platform_version.split('.')[0]
+        multilib_artifacts = []
+
+        # FIXME: Hardcoded platform for now, should send actual platform
+        #  in the task ref
+        multilib_packages = await MultilibProcessor.get_module_multilib_data(
+            beholder_client, 'AlmaLinux', major_distr_version,
```
anfimovdm, line 155: We can use platform data from `self._platforms`, i'm think there also should be `distr_version`

alws/build_planner.py
```diff
@@ -230,16 +283,10 @@ async def add_task(self, task: build_schema.BuildTaskRef):
...
                 module_index.add_module(module)
-                if len(module_templates) > 1:
-                    devel_module = ModuleWrapper.from_template(
-                        module_templates[1],
-                        module.name + '-devel',
-                        module.stream
-                    )
+                if module_index.has_devel_module():
+                    devel_module = module_index.get_module(
+                        f'{task.module_name}', task.module_stream)
```
anfimovdm, line 289: does we need here `-devel` suffix? or it's works fine for devel modules?

anfimovdm (changes requested):

alws/build_planner.py
Korulag: Yep, you're right, missed it

alws/build_planner.py
Korulag: Makes sense, will fix the code

alws/build_planner.py
```diff
@@ -135,16 +140,88 @@ async def init_build_repos(self):
...
+            has_devel: bool = False
+    ) -> typing.List[dict]:
+        beholder_client = BeholderClient(
+            settings.beholder_host, token=settings.beholder_token)
+        multilib_artifacts = []
+
+        # FIXME: Hardcoded platform for now, should send actual platform
+        #  in the task ref
+        platform = self._platforms[0]
+        platform_name = platform.name,
```
anfimovdm, line 154: comma

alws/build_planner.py
```diff
@@ -135,16 +140,88 @@ async def init_build_repos(self):
...
+            self, task: build_schema.BuildTaskModuleRef,
+            has_devel: bool = False
+    ) -> typing.List[dict]:
+        beholder_client = BeholderClient(
+            settings.beholder_host, token=settings.beholder_token)
+        multilib_artifacts = []
+
+        # FIXME: Hardcoded platform for now, should send actual platform
+        #  in the task ref
+        platform = self._platforms[0]
```
anfimovdm, line 153: in case for build packages/modules for few platforms, it would be wrong getting only first platform

anfimovdm, line 153: if we would build packages/modules for few platforms, there always be only first platform in list
for `platform_name` it would be ok, but not for `major_distr_version`

alws/build_planner.py
```diff
@@ -135,16 +140,88 @@ async def init_build_repos(self):
...
+    async def get_multilib_artifacts(
+            self, task: build_schema.BuildTaskModuleRef,
+            has_devel: bool = False
+    ) -> typing.List[dict]:
+        beholder_client = BeholderClient(
+            settings.beholder_host, token=settings.beholder_token)
+        multilib_artifacts = []
+
+        # FIXME: Hardcoded platform for now, should send actual platform
+        #  in the task ref
```
anfimovdm, line 152: this `fixme` can be deleted

anfimovdm (changes requested):

soksanichenko (approved):

PR #201, opened 2022-05-19
Fixes duplicated module release

anfimovdm:


alws/release_planner.py
```diff
@@ -215,10 +215,10 @@ async def prepare_data_for_executing_async_tasks(
             for repo in self.base_platform.repos:
                 pulp_repo_info = pulp_repos[repo.pulp_href]
                 self.repo_data_by_href[repo.pulp_href] = (repo.id, repo.arch)
-                is_debug = bool(re.search(r'debug(info|source|)',
-                                          pulp_repo_info['name']))
+                repo_is_debug = bool(re.search(r'debug(info|source|)',
```
soksanichenko, line 218: `re.search(...) is not None`

soksanichenko (approved):

alws/release_planner.py
AvoidMe, line 218: Please, remove last `|`, since `name-debug` package is not a debug package.

alws/utils/modularity.py
```diff
@@ -301,6 +301,10 @@ def iter_dependencies(self):
                 for stream in dep.get_buildtime_streams(module):
                     yield module, stream
 
+    def get_nsvca(self) -> str:
+        return (f'{self.name}:{self.stream}:{self.version}:'
+                f'{self.context}:{self.arch}')
```
AvoidMe, line 306: Please, make it property, just like the rest of attributes (name, stream, version, etc)

alws/release_planner.py
anfimovdm: we looking here for debug repos, not packages, but i'll fix it anyway

AvoidMe: Sorry, was confused by previous change, than last `|` should stay

AvoidMe (approved):

PR #209, opened 2022-05-24
Added script to delete packages from repository

Korulag:


scripts/delete_from_repository.py
```diff
@@ -0,0 +1,144 @@
...
+        if not repo:
+            raise RepositoryNotFound(f'Repository {repo_name} is not found')
+        return repo
+
+
+def wait_for_task(base_url: str, task_href: str, auth: HTTPBasicAuth):
+    task_url = urljoin(base_url, task_href)
+    res = requests.get(task_url, auth=auth).json()
+    while res['state'] not in ('failed', 'completed'):
+        time.sleep(3)
```
soksanichenko, line 51: Add some print/logging which shows that the script doesn't stuck

scripts/delete_from_repository.py
```diff
@@ -0,0 +1,144 @@
...
+        urljoin(pulp_host, packages_endpoint), params=query_params,
+        auth=pulp_auth).json()
+    package_count = pulp_response['count']
+    repo_packages = pulp_response['results']
+    logger.debug('Packages count: %d', package_count)
+    if package_count > 100:
+        query_params['limit'] = package_count + 1
+        repo_packages = requests.get(
+            urljoin(pulp_host, packages_endpoint),
+            params=query_params, auth=pulp_auth).json()['results']
```
anfimovdm, line 105: we rewrite here results from prev responses, we should extend list instead
also, increasing limit by one is kinda strange, it seems like slow us down
u can look on `iter_repo` func in `alws/utils/uploader.py:18` how we get all content from repo

scripts/delete_from_repository.py
```diff
@@ -0,0 +1,144 @@
...
+                to_remove.append(pulp_pkg['pulp_href'])
+    logger.info('List is compiled')
+
+    if not to_remove:
+        logger.warning('Nothing to delete, packages %s'
+                       ' were not found in the repository',
+                       str(arguments.packages))
+        return
+
+    modify_url = urljoin(urljoin(pulp_host, repo.pulp_href), 'modify/')
```
anfimovdm, line 126: we already make this repo url above `urljoin(pulp_host, repo.pulp_href)`, we can reuse it

anfimovdm (changes requested):

anfimovdm (approved):

scripts/delete_from_repository.py
```diff
@@ -0,0 +1,151 @@
...
+
+def get_repository(repo_name: str, arch: str) -> models.Repository:
+    with SyncSession() as session:
+        repo = (session.execute(select(models.Repository)
+                                .where(models.Repository.name == repo_name,
+                                       models.Repository.arch == arch))
+                .scalars().first())
+        if not repo:
+            raise RepositoryNotFound(f'Repository {repo_name} is not found')
+        return repo
```
AvoidMe, line 44: How will package maintainers get access to our database?
IMHO, it's a lot of unneeded dependencies for this task

AvoidMe: Basically, it's a lot easier to get repo from pulp just by name: https://docs.pulpproject.org/pulp_rpm/restapi.html#operation/repositories_rpm_rpm_list
Also, from pulp response (autopublish) you can decide should or shouldn't create publication manually, it's a lot safer / require a lot less dependencies / can be runned on any device any time

Korulag: Do you mean we need to rely only on communication with pulp?

AvoidMe: Yes, this is how I did it previously, I can share my work with you
My script rely only on pulp responses and is safe to run from any place any time

scripts/delete_from_repository.py
```diff
@@ -0,0 +1,152 @@
...
+    full_url = urljoin(base_url, '/pulp/api/v3/publications/rpm/rpm/')
+    task_href = requests.post(
+        full_url, json={'repository': repo_href}, auth=auth).json()['task']
+    return wait_for_task(base_url, task_href, auth, logger)
+
+
+def get_repository(base_url: str, repo_name: str, arch: str,
+                   auth: HTTPBasicAuth) -> typing.Optional[dict]:
+    full_url = urljoin(base_url, '/pulp/api/v3/repositories/rpm/rpm/')
+    params = {'name__contains': f'{repo_name}-{arch}'}
```
anfimovdm, line 54: why we can't search repo by exact name?
it's looks unsafe search repo without arch, we can get wrong repo

anfimovdm (changes requested):

scripts/delete_from_repository.py
Korulag: Pulp does not have info on the repository architecture

Korulag: Fixed a search though, to ensure we pick up the correct repo

anfimovdm (approved):

PR #210, opened 2022-05-25
Fix signature verification

Korulag:


scripts/packages_exporter.py
```diff
@@ -348,59 +377,62 @@ async def check_noarch_in_user_distribution_repos(self, distr_name: str):
...
-                        break
-                if not signed_with_subkey:
-                    self.logger.error('Package %s is signed with wrong key, '
-                                      'expected "%s", got "%s"',
-                                      package_path, str(key_ids_lower),
-                                      pkg_key_id)
-                    wrong_signature_packages.add(f'{package_path} {pkg_key_id}')
+        futures = {}
+
+        with ThreadPoolExecutor(max_workers=10) as executor:
```
anfimovdm, line 416: can we use here `asyncio.gather`?
i may be wrong, but it's kinda strange or maybe wrong to call `ThreadPoolExecutor` inside `async` func that have eventloop and etc.

scripts/packages_exporter.py
```diff
@@ -33,6 +35,13 @@
     '~/config/known_subkeys.json'))
 
 
+class SignStatusEnum(IntEnum):
+    SUCCESS = 1
+    READ_ERROR = 2
+    NO_SIGNATURE = 3
+    WRONG_SIGNATURE = 4
```
anfimovdm, line 42: need to be moved in constants

anfimovdm, line 42: need to be moved in constants, not critical

anfimovdm (changes requested):

scripts/packages_exporter.py
Korulag: The function for checking signatures is synchronous, RPM bindings are also synchronous, I've tested this approach and everything is working fine with this setup. Calling synchronous code in async function is not prohibited, it's just locking the context until function is completed

anfimovdm (approved):

soksanichenko (approved):

PR #211, opened 2022-05-25
ALBS-395: Added repos for AlmaLinux 9

anfimovdm:


reference_data/platforms.yaml
```diff
@@ -809,72 +809,102 @@
     - arch: s390x
       name: almalinux-9-devel
       type: rpm
-      remote_url: https://repo.almalinux.org/development/almalinux/9/devel/s390x/
-      production: false
+      remote_url: http://rsync.repo.almalinux.org/almalinux/9/devel/s390x/os/
+      export_path: almalinux/9/devel/s390x/os/
+      production: true
       debug: false
+      remote_sync_policy: on_demand
```
Korulag, line 816: Since usually we use 'immediate' and 'mirror_complete' we might as well make it a default value at least for Alma 9

Korulag (changes requested):

Korulag (approved):

soksanichenko (approved):

PR #221, opened 2022-06-14
Moved completing test tasks into dramatiq queue

Korulag:


alws/dramatiq/tests.py
```diff
@@ -0,0 +1,34 @@
...
+        except Exception as e:
+            logging.exception(
+                'Cannot set test task "%d" result, marking as failed.'
+                'Error: %s', task_id, str(e))
+            await db.rollback()
+        await db.execute(update(TestTask).where(
+            TestTask.id == task_id).values(status=TestTaskStatus.FAILED))
+
+
+@dramatiq.actor(max_retries=2, priority=1, time_limit=DRAMATIQ_TASK_TIMEOUT)
```
AvoidMe, line 31: Please, set a priority for higher value, sign tasks and build done also have priority=1, this will make impossible for build to complete or sign if test system is under load

alws/crud/test.py
```diff
@@ -103,42 +104,41 @@ async def complete_test_task(db: Session, task_id: int,
...
+        elif key == 'logs':
+            continue
+        elif not item.get('success', False):
+            status = TestTaskStatus.FAILED
+            break
+    task.status = status
+    task.alts_response = test_result.dict()
+    db.add(task)
+    db.add_all(logs)
+    await db.flush()
```
AvoidMe, line 141: probably unnecessary flush, since we have a transaction

alws/dramatiq/tests.py
Korulag: Fixed

alws/crud/test.py
Korulag: Removed the flush

PR #228, opened 2022-06-21
ALBS-314: Set BuildTask error field for fast-failed tasks

javihernandez:
This way we produce a user facing error in frontend indicating why the
task was marked as failed.

Example:

image

### DIFF
--- a/alws/crud/build_node.py
+++ b/alws/crud/build_node.py
@@ -471,13 +471,21 @@ async def __update_built_srpm_url(db: Session, build_task: models.BuildTask):
     )
     srpm_artifact = srpm_artifact.scalars().first()
 
-    # if SRPM isn't built in first arch of project,
-    # we need to stop building project
+    # if SRPM isn't built in first arch of project, we need to stop building
+    # the project and fast-fail the uncompleted tasks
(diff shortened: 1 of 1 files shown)


alws/crud/build_node.py
```diff
@@ -471,15 +471,26 @@ async def __update_built_srpm_url(db: Session, build_task: models.BuildTask):
...
 
-    # if SRPM isn't built in first arch of project,
-    # we need to stop building project
+    # if SRPM isn't built in first arch of project, we need to stop building
+    # the project and fast-fail the uncompleted tasks
     if (not srpm_artifact and build_task.status == BuildTaskStatus.FAILED
             and uncompleted_tasks_ids):
         update_query = update(models.BuildTask).where(
             models.BuildTask.id.in_(uncompleted_tasks_ids),
         ).values(status=BuildTaskStatus.FAILED)
```
anfimovdm, line 480: we can update "fast-fail log" here too, instead of additional update query

anfimovdm (changes requested):

alws/crud/build_node.py
javihernandez: Nice catch!

Korulag (approved):

PR #244, opened 2022-07-01
Cas release integration

anfimovdm:


Dockerfile
```diff
@@ -1,9 +1,10 @@
 FROM fedora:34
 
+RUN curl https://packages.codenotary.org/codenotary.repo -o /etc/yum.repos.d/codenotary.repo
 RUN mkdir -p /code && \
     yum update -y && \
-    yum install python3-virtualenv python39 libmodulemd python3-libmodulemd \
-                python3-libmodulemd1 modulemd-tools python-gobject -y && \
+    yum install --enablerepo="codenotary-repo" cas python3-virtualenv python39 libmodulemd python3-libmodulemd \
+                python3-libmodulemd1 modulemd-tools python-gobject git -y && \
```
javihernandez, line 7: Question. Do we need git?

javihernandez (approved):
I only left a question regarding the need to install git in the container. Other than that, and to my little understanding of the system, LGTM.

Dockerfile
anfimovdm: yeah, we need to install `git` for installing `cas_wrapper`

soksanichenko (approved):

PR #245, opened 2022-07-05
Added initial RBAC implementation

Korulag:
This commit brings initial implementation of the RBAC in the build
system. This will allow to manage permissions over different objects
using user groups.


alws/models.py
```diff
@@ -502,6 +568,116 @@ class BinaryRpm(Base):
     source_rpm = relationship('SourceRpm', back_populates='binary_rpms')
 
 
+class UserAction(Base):
+
+    __tablename__ = 'user_actions'
+
+    id = sqlalchemy.Column(sqlalchemy.Integer, primary_key=True)
+    name = sqlalchemy.Column(sqlalchemy.String(100), unique=True)
```
anfimovdm, line 576: may be this column should have text type instead of string?

anfimovdm (changes requested):

alws/perms/actions.py
```diff
@@ -0,0 +1,198 @@
...
+)
+
+# Team actions
+CreateTeam = Action(
+    name='create_team',
+    description='Ability to create a team'
+)
+ReadTeam = Action(
+    name='read_team',
+    description='Ability to create a team'
```
javihernandez, line 71: Ability to read

alws/perms/actions.py
```diff
@@ -0,0 +1,198 @@
...
+    name='create_platform',
+    description='Ability to create a platform'
+)
+ReadPlatform = Action(
+    name='read_platform',
+    description='Ability to read a platform information'
+)
+UpdatePlatform = Action(
+    name='update_platform',
+    description='Ability to update platform'
```
javihernandez, line 131: Ability to update **a** platform

alws/models.py
```diff
@@ -4,15 +4,87 @@
...
+    permissions = sqlalchemy.Column(sqlalchemy.Integer, nullable=False,
+                                    default=764)
+
+    @property
+    def permissions_triad(self):
+        self.validate_permissions(self.permissions)
+        return PermissionTriad(
+            Permissions(self.permissions // 100),
+            Permissions(self.permissions // 10 % 10),
+            Permissions(self.permissions % 10),
```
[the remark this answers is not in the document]
AvoidMe: We already has all of the values in `validate_permissions`, may be we could return tuple `(owner, group, other)` from here? Because this code is a little obfuscated

alws/models.py
```diff
@@ -222,12 +294,12 @@ class Repository(CustomRepoRepr):
...
     __tablename__ = 'repository_remotes'
     __tableargs__ = [
-        sqlalchemy.UniqueConstraint('name', 'arch', 'url')
+        sqlalchemy.UniqueConstraint('name', 'arch', 'url', name='repo_remote_uix')
     ]
 
     id = sqlalchemy.Column(sqlalchemy.Integer, primary_key=True)
     name = sqlalchemy.Column(sqlalchemy.Text, nullable=False)
-    arch = sqlalchemy.Column(sqlalchemy.Text, nullable=False)
+    arch = sqlalchemy.Column(sqlalchemy.Text, nullable=True)
```
AvoidMe, line 302: In which case arch for repository can be null?

scripts/bootstrap_permissions.py
```diff
@@ -0,0 +1,98 @@
...
+
+            db.add_all(objs)
+            await db.commit()
+        except Exception:
+            logger.exception('Cannot do a bootstrap of the permissions')
+            await db.rollback()
+
+
+if __name__ == '__main__':
+    sync(main())
```
AvoidMe, line 98: main function is async already, what's the point in `sync()`, instead of `asyncio.run()`?

scripts/bootstrap_permissions.py
```diff
@@ -0,0 +1,98 @@
...
+            for user in existing_users:
+                user.teams.append(alma_team)
+                user.roles.append(contributor_role)
+                objs.append(user)
+            for sign_key in existing_sign_keys:
+                sign_key.roles.append(signer_role)
+                objs.append(sign_key)
+
+            db.add_all(objs)
+            await db.commit()
```
AvoidMe, line 91: No need to manually call `commit()` and `rollback()` inside a `with db.begin()`
Also, you can just remove this `try/except` end everything will be working as before, almost no changes will be.

alws/models.py
```diff
@@ -4,15 +4,87 @@
...
+    @declared_attr
+    def owner(cls):
+        return relationship('User')
+
+    permissions = sqlalchemy.Column(sqlalchemy.Integer, nullable=False,
+                                    default=764)
+
+    @property
+    def permissions_triad(self):
+        self.validate_permissions(self.permissions)
```
AvoidMe, line 70: Why there is validate on get, but no validation on set, I think, it could be more appropriate on set

AvoidMe (changes requested):

alws/models.py
Korulag: For now we do not set permissions anywhere, only default values that are correct

anfimovdm (approved):

PR #248, opened 2022-07-14
ALBS-409: Wait for other build tasks before creating test tasks

javihernandez:


javihernandez:
Questions:
* Not sure whether that's the best place to do it - maybe better doing it [here](https://github.com/AlmaLinux/albs-web-server/blob/af9fc0e7f00a9c30f54aa97fd18585f6df4a2a95/alws/dramatiq/build.py#L56-L60)?
* Not sure if it's okay to wait the way we're doing right now
* How about using the `get_db` approach? Can I reuse the session I already get in [here](https://github.com/AlmaLinux/albs-web-server/blob/af9fc0e7f00a9c30f54aa97fd18585f6df4a2a95/alws/crud/test.py#L19)?

anfimovdm (approved):

alws/dramatiq/build.py
```diff
@@ -65,6 +73,50 @@ async def _create_log_repo(task_id: int):
...
+    async with db.begin():
+        build_id, platform_id = (await db.execute(select(
+            models.BuildTask.build_id,
+            models.BuildTask.platform_id,
+        ).where(
+            models.BuildTask.id == build_task_id
+        ))).first()
+
+        all_completed = await _check_build_and_completed_tasks(db, build_id, platform_id)
+        while not all_completed:
```
Korulag, line 113: This one will hold function possibly for a very long time. Probably we just need to do one-time check and if not all tasks are completed then just skip creating test tasks

Korulag (changes requested):

alws/crud/test.py
```diff
@@ -3,18 +3,34 @@
...
 from alws.utils.pulp_client import PulpClient
 from alws.utils.file_utils import download_file
 from alws.utils.parsing import parse_tap_output, tap_set_status
 
+async def create_test_tasks_for_build_id(db: Session, build_id: int):
+    async with db.begin():
+        ## We get all build_tasks with the same build_id
+        ## and whose status is COMPLETED
+        build_task_ids = (await db.execute(
+            select(models.BuildTask.id).where(and_(
```
Korulag, line 25: `and` is default grouping of conditions in SQLAlchemy, so you can skip it in simple requests

alws/dramatiq/build.py
```diff
@@ -66,6 +77,45 @@ async def _create_log_repo(task_id: int):
...
+    async with db.begin():
+        build_tasks = (await db.execute(
+            select(func.count()).select_from(models.BuildTask).where(
+                models.BuildTask.build_id == build_id
+            )
+        )).scalar()
+
+        completed_tasks = (await db.execute(
+            select(func.count()).select_from(models.BuildTask).where(and_(
+                models.BuildTask.build_id == build_id,
```
Korulag, line 101: Same here, no need in `and_` operation

Korulag (changes requested):

Korulag (approved):

anfimovdm (approved):

PR #251, opened 2022-07-19
Integrate fastapi users

Korulag:
This commits brings integration with `fastapi-users` library which allows for easier integration with different IDPs.


alws/routers/builds.py
```diff
@@ -35,14 +36,18 @@
...
         ):
-    return await build_crud.create_build(
-        db,
-        build,
-        user['identity']['user_id']
-    )
+    # `get_current_user returns tuple of user record and token,
+    # so to address user record we need to pick first item in `user_data`
+    try:
+        return await build_crud.create_build(db, build, user.id)
```
anfimovdm, line 45: in comment above you mention that `get_current_user` returns tuple, but you use `id` attribute here

alws/models.py
```diff
@@ -685,21 +694,52 @@ def __repr__(self):
...
+
+    id = sqlalchemy.Column(sqlalchemy.Integer, primary_key=True)
+
+    @declared_attr
+    def user_id(cls):
+        return sqlalchemy.Column(
+            sqlalchemy.Integer,
+            sqlalchemy.ForeignKey("users.id", ondelete="cascade"),
+            nullable=False
+        )
```
anfimovdm, line 708: I'm not sure but may be we should add `relationship` field for `User` model
or this field already inherited from `SQLAlchemyBaseOAuthAccountTable`?

alws/models.py
```diff
@@ -685,21 +694,52 @@ def __repr__(self):
...
+    id = sqlalchemy.Column(sqlalchemy.Integer, primary_key=True,
+                           autoincrement=True)
+
+    @declared_attr
+    def user_id(cls):
+        return sqlalchemy.Column(
+            sqlalchemy.Integer,
+            sqlalchemy.ForeignKey("users.id", ondelete="cascade"),
+            nullable=False
+        )
```
anfimovdm, line 723: same as above

alws/routers/builds.py
```diff
@@ -8,22 +8,23 @@
...
+from alws.dependencies import get_db
+from alws.errors import DataNotFoundError, PermissionDenied
 from alws.schemas import build_schema
 
 
 router = APIRouter(
     prefix='/builds',
     tags=['builds'],
-    dependencies=[Depends(JWTBearer())]
+    dependencies=[Depends(get_current_user)]
```
anfimovdm, line 27: should we invoke `get_current_user` here like `JWTBearer`?

anfimovdm (changes requested):

alws/routers/builds.py
Korulag: Yeah, didn't remove the comment, it's actually returning a User object

alws/models.py
Korulag: This field is required in base class, so since we're using a different table name I needed to patch foreignkey. And yes, relationship is already defined in base class

alws/models.py
Korulag: Same as above :)

alws/routers/builds.py
Korulag: Nope, it's ready to use without invocation

anfimovdm (approved):

PR #255, opened 2022-07-25
Added fucntionality to activate and deactivate users

Korulag:


alws/models.py
```diff
@@ -733,6 +733,8 @@ class User(SQLAlchemyBaseUserTable[int], Base):
     last_name = sqlalchemy.Column(sqlalchemy.String(320), nullable=True)
     hashed_password: str = sqlalchemy.Column(
         sqlalchemy.String(length=1024), nullable=True)
+    is_active: bool = sqlalchemy.Column(
```
anfimovdm, line 736: not sure what we really need here type hint

anfimovdm (approved):

alws/models.py
Korulag: copy-paste :)

PR #260, opened 2022-08-05
Fixed regression in module multilib artifacts placement

Korulag:


alws/utils/multilib.py
```diff
@@ -249,9 +249,15 @@ async def update_module_index(module_index: IndexWrapper, module_name: str,
                                   module_stream: str, packages: list):
         if not packages:
             return
-        if module_index.has_devel_module() and 'python' not in module_name:
-            module = module_index.get_module(
-                f'{module_name}-devel', module_stream)
+        # Basically now we always will have devel module,
+        # but in case if not the original behavior is saved
+        if module_index.has_devel_module():
```
AvoidMe, line 254: I think, it's useless check now, we can safely remove it

Korulag: Fixed

alws/utils/multilib.py
```diff
@@ -249,11 +249,13 @@ async def update_module_index(module_index: IndexWrapper, module_name: str,
                                   module_stream: str, packages: list):
         if not packages:
             return
-        if module_index.has_devel_module() and 'python' not in module_name:
+
+        # Ruby and Python modules should have multilib in their main module
+        if 'ruby' in module_name or 'python' in module_name:
```
anfimovdm, line 254: I'm think that main problem here, because we have in index usual and devel model, can we process modules successively?

anfimovdm (changes requested):

alws/utils/multilib.py
Korulag: Yes, before having devel modules for only a handful of modules logic was correct. But now we have devel modules for all and multilib for ruby started acting out. I've put a specific condition for ruby to move multilib into the main module

anfimovdm (approved):

AvoidMe (approved):

PR #263, opened 2022-08-09
Fixed integrity error, added logs in build_done

anfimovdm:
`IntegrityError` raises in case, if we try to process noarch packages in latest build task in line and when this build task have status `EXCLUDED`

### DIFF
--- a/alws/crud/build_node.py
+++ b/alws/crud/build_node.py
@@ -243,7 +243,7 @@ async def __process_rpms(db: Session, pulp_client: PulpClient, task_id: int,
             raise RepositoryAddError(
                 f'Cannot add RPM packages to the repository {str(repo)}'
             )
-    
+
     def append_errata_package(_, errata_package, artifact, rpm_info):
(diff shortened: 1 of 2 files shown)


alws/crud/build_node.py
```diff
@@ -637,6 +648,10 @@ async def build_done(db: Session, pulp: PulpClient,
             srpm = db_srpm.scalars().first()
             if build_task.built_srpm_url is not None and srpm is not None:
                 continue
+        # if build task is excluded or failed, we don't need to create
+        # SourceRpm and BinaryRpm records, because it's break package releases
```
javihernandez, line 652: 'because it breaks'

javihernandez (approved):

maccelf (approved):

soksanichenko (approved):

alws/crud/build_node.py
anfimovdm: Thanks, fixed

PR #275, opened 2022-08-18
Fixes for CVE`s titles and for missing packages in bulletins

anfimovdm:


scripts/errata_fix_script.py
```diff
@@ -0,0 +1,199 @@
...
+async def update_albs_db():
+    logging.info("Update albs db started")
+    query = (
+        select(ErrataReference)
+        .where(
+            ErrataReference.ref_type != ErrataReferenceType.bugzilla,
+            ErrataReference.title == '',
+        )
+    )
+    async for db in get_db():
```
Korulag, line 182: you can use `contextlib.asynccontextmanager(get_db)()` construct, this way you'll be able to write whole `with` statement in one string

soksanichenko (approved):
will lgtm when you fix the Vasily's comment

scripts/errata_fix_script.py
```diff
@@ -0,0 +1,199 @@
...
+                pulp_pkg = albs_packages_cache[filename]
+                if (
+                    pulp_pkg['sha256'] is None
+                    and pulp_pkg['checksum_type'] == 'sha256'
+                ):
+                    logging.warning(
+                        'sha256 sum is missing for %s using pkgId instead',
+                        filename,
+                    )
+                    pulp_pkg['sha256'] = pulp_pkg['pkgId']
```
javihernandez, line 133: Maybe stupid question. Is this the right thing to do here? Why not just leaving it empty?

javihernandez (commented):
LGTM and will definitely approve it when comments/questions are addressed/responded. Great job!!

scripts/errata_fix_script.py
anfimovdm: this field have not null constraint in pulp DB

javihernandez (approved):

Korulag (approved):

PR #277, opened 2022-08-22
Fixed released status in builds after release

anfimovdm:


Korulag (approved):

alws/release_planner.py
```diff
@@ -1177,9 +1177,15 @@ async def commit_release(
                 message = 'Successfully committed release'
                 message += '\n'.join(release_messages)
                 release.status = ReleaseStatus.COMPLETED
+                release_status = True
+            for build in release_builds.scalars().all():
```
Korulag, line 1181: It's probably better to do `update` statement instead of this cycle. Basically you need to set `release_id` field on the builds to the release + set the status, something like `update().values(release_id=release.id, status=release_status)`

Korulag (changes requested):

alws/release_planner.py
anfimovdm: Fixed

maccelf (approved):

javihernandez (approved):

Korulag (approved):

PR #281, opened 2022-08-25
Added script to copy packages from one Pulp repo to another

Korulag:


javihernandez (commented):
LGTM, however. I've got a couple of questions
1. What if we want to copy a considerable number of packages? Or even all of them? 
2. What happens if a package already exists in the target repo? i.e.: I pass a package that already exists in my target repo by mistake

Korulag:
1. Since "packages in the repository" inside pulp is basically a relationship between tables, adding even considerable amount of packages should be okay, this may affect time of execution, but not greatly, since the most time-consuming task is metadata update.
2. Basically package either will be overwritten or it's the same package as was already in a repo (by id) and it does not affect repository at all. I wanted this logic to be as KISS as possible, so all the blame on the person who runs the script :)

soksanichenko (approved):
LGTM

PR #286, opened 2022-08-30
Added new team creation when product is created

Korulag:


alws/crud/products.py
```diff
@@ -50,12 +51,21 @@ async def create_product(
...
+    team_name = f'{payload.name}_team'
+    existing_team = await get_teams(db, name=team_name)
+    if existing_team:
+        raise ProductError(
+            f"Product's team name intersects with the existing team, "
+            f"which may lead to permissions issues")
 
-    product = models.Product(**payload.dict())
+    team_payload = TeamCreate(team_name=team_name, user_id=payload.owner_id)
+    team_roles = await create_team_roles(db, team_name)
```
anfimovdm, line 62: it's unnecessary, we create team_roles in `create_team` func

alws/crud/products.py
```diff
@@ -50,12 +51,21 @@ async def create_product(
     if product:
         raise ProductError(f'Product with name={payload.name} already exist')
 
-    if not can_perform(team, owner, actions.CreateProduct.name):
-        raise PermissionDenied(
-            'User has no permissions to create the product'
-        )
+    team_name = f'{payload.name}_team'
```
anfimovdm, line 54: may be we should use `payload.name.lower()`?

anfimovdm (changes requested):

alws/crud/products.py
javihernandez: right, see https://github.com/AlmaLinux/albs-web-server/blob/bc6521061c9b9106f73f5e7f67649b2a72a2b2c9/alws/crud/teams.py#L100

Korulag: It's just easier to get a list of roles this way to add them later to the user. I know that it's built-in for `create_team`, but wanted to make code clearer when it comes to SQL queries.

javihernandez: Got it, and `create_team_roles` checks if the roles are already there. Ok for me.

javihernandez (approved):

anfimovdm (approved):

PR #289, opened 2022-08-30
ALBS-595: Added backend support of admin user page

javihernandez:


soksanichenko (approved):
lgtm

soksanichenko (approved):
The last changes lgtm

alws/crud/user.py
```diff
@@ -34,18 +40,227 @@ async def get_all_users(db: Session) -> typing.List[models.User]:
 async def activate_user(user_id: int, db: Session):
     await db.execute(update(models.User).where(
         models.User.id == user_id).values(is_verified=True, is_active=True))
+    await db.commit()
```
anfimovdm, line 43: Adding here `db.commit` will affect us in `scripts/manage_users.py`, it also related for `deactivate_user`, `make_superuser` and `make_usual_user`

anfimovdm (changes requested):

alws/crud/user.py
javihernandez: You're right @anfimovdm. In fact, this is a leftover from where I first started working on this. Initially, I was going to reuse these methods but in the end I'm not, so all these `db.commit` can go away without any problem. I'll create a ticket to update the `manage_users.py` script and get rid of these functions - as they are only used by that script.

javihernandez: Fixed in https://github.com/AlmaLinux/albs-web-server/pull/289/commits/cdbf976192f067970f7de0bcc55e2ae067860ce0

anfimovdm (approved):

PR #291, opened 2022-08-31
Fixed showing last logs in release

anfimovdm:


alws/release_planner.py
```diff
@@ -395,7 +395,9 @@ async def commit_release(
 
         builds_released = False
         try:
-            release_messages = await self.execute_release_plan(release)
+            release_messages, release_plan = await self.execute_release_plan(
+                release,
+            )
```
Korulag, line 400: Can we just make `message` variable here like `'\n'.join(release_messages)` and skip all other changes? I just do not understand a point of returning release plan from the execution function

Korulag (changes requested):

alws/release_planner.py
anfimovdm: Good point, I'll check if we modifying release plan or not during `execute_release_plan`

anfimovdm: As I thought, we modify release plan during execution
https://github.com/AlmaLinux/albs-web-server/blob/9fb851dc25611fa2c3a7c98e3464f7ced31055df/alws/release_planner.py#L1245-L1252
We can send copy of release plan into `execute_release_plan` like it was before or just use current approach

Korulag (approved):

maccelf (approved):

PR #303, opened 2022-09-08
Added minor improvements for releases

anfimovdm:
I have some doubts about `func.current_timestamp()`, I think it depends on timezone in DB host, **we need to check it**


soksanichenko (approved):

alws/crud/release.py
```diff
@@ -19,24 +19,57 @@
...
+        ).order_by(models.Release.id.desc())
+        if count:
+            query = select(func.count(models.Release.id))
+        if release_id:
+            query = query.where(models.Release.id == release_id)
+        if search_params:
+            # These links could be helpful for filtering by JSON
+            # https://github.com/sqlalchemy/sqlalchemy/discussions/7991
+            # https://www.postgresql.org/docs/9.5/functions-json.html
+            # if search_params.package_name:
```
soksanichenko, line 42: Should it be commented?

soksanichenko (commented):
дпеь

javihernandez:
> I have some doubts about `func.current_timestamp()`, I think it depends on timezone in DB host, **we need to check it**

Right, sqlalchemy's `current_timestamp` calls Postgresql's [CURRENT_TIMESTAMP](https://www.postgresqltutorial.com/postgresql-date-functions/postgresql-current_timestamp/), which includes the timezone offset.

alws/crud/release.py
anfimovdm: I comment it by purpose, I don't have enough time right now for researching how to make a query for JSON field, so I provide some links for the future

alws/alembic/versions/76855243a038_added_timestamp_for_releases.py
```diff
@@ -0,0 +1,34 @@
...
+revision = '76855243a038'
+down_revision = 'ec505e94fd1a'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('build_releases', sa.Column('created_at', sa.DateTime(), nullable=True))
+    op.alter_column('build_releases', 'product_id',
```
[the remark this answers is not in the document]
anfimovdm: I'm not sure, but it appears after population `Product` entities for releases, at first it was nullable, but after it became not nullable

javihernandez (approved):

maccelf (approved):

soksanichenko (approved):

javihernandez (approved):

PR #308, opened 2022-09-13
ALBS-638: Added models for SBOM CycloneDX formatter

javihernandez:
Also, added a script that uses these models


javihernandez:
Moved elsewhere

PR #310, opened 2022-09-15
Fixed approach for restoring removed packages

anfimovdm:


maccelf (approved):

soksanichenko (approved):

alws/utils/uploader.py
```diff
@@ -77,13 +74,34 @@ async def upload_modules(self, repo_href: str,
...
         # need to remove them before creating new modulemd
         await self.pulp.modify_repository(repo_href, remove=modules_to_remove)
+        logging.debug(
+            'Removed the following entities from repo "%s":\n%s',
+            repo_href,
+            modules_to_remove,
+        )
         # if we fall during next modifying repository, we can delete this
         # repo version and rollback all changes that makes upload
         new_version_href = await self.pulp.get_repo_latest_version(repo_href)
```
Korulag, line 88: I think this may be a delayed race condition. Consider following: you're doing modification and some other task does modification and it will be finished later than you/your task. Then you may pick up the version from that task, not yours. I think we need to get repository version reference from task result of repository modification

Korulag (changes requested):

alws/utils/uploader.py
anfimovdm: Thanks, fixed

javihernandez (approved):

PR #311, opened 2022-09-16
Errata release improvements

anfimovdm:


alws/crud/errata.py
```diff
@@ -913,20 +914,21 @@ async def prepare_updateinfo_mapping(
...
+                models.BuildTaskArtifact.href == pkg_href,
             ).options(
                 selectinload(models.BuildTaskArtifact.build_task)
                 .selectinload(models.BuildTask.rpm_module)
             )
         )).scalars().all()
         for db_pkg in db_pkg_list:
             errata_pkgs = await db.execute(
-                select(models.ErrataToALBSPackage).where(
+                select(models.ErrataToALBSPackage).where(or_(
```
javihernandez, line 928: What is the point of `or_` here?

maccelf (approved):

javihernandez (commented):
Aside the minor comment I left, LGTM

alws/crud/errata.py
anfimovdm: When we're creating new errata record, we looking for packages in the platform repositories and in our build artifacts
We need here `or_` in case when we add new errata record and some packages needed for errata already released, we use `pulp_href` field instead of `albs_artifact_id` field
(see `search_for_albs_packages()`)

javihernandez (approved):

PR #315, opened 2022-09-28
ALBS-546: asio.src placed to mariadb module but should be in mariadb-devel

soksanichenko:
- We should add a source RPM into a non-devel module if a source RPM produced
   at least one RPM. Otherwise, we should add that source RPM into a devel module.

### DIFF
--- a/alws/crud/build_node.py
+++ b/alws/crud/build_node.py
@@ -303,18 +303,12 @@ async def __process_rpms(db: Session, pulp_client: PulpClient, task_id: int,
         logging.info('RPM package: %s', json.dumps(rpm_package, indent=4))
         logging.info('Packages info: %s', json.dumps(packages_info, indent=4))
         logging.info('Module: %s is devel: %s', module.name, module.is_devel)
-        if rpm_package['rpm_sourcerpm']:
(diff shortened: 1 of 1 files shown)


maccelf (approved):

alws/crud/build_node.py
```diff
@@ -298,8 +300,41 @@ def append_errata_package(_, errata_package, artifact, rpm_info):
             )
         rpms.append(artifact)
 
+    def is_source_rpm_included(
```
Korulag, line 303: Since the function works not only with SRPMs, maybe change it's name accordingly?

alws/crud/build_node.py
```diff
@@ -298,8 +300,41 @@ def append_errata_package(_, errata_package, artifact, rpm_info):
...
+            rpm_pkg: dict,
+            pgks_info: dict,
+            current_module: ModuleWrapper
+    ) -> bool:
+        source_rpm_list = [
```
Korulag, line 308: This will be compiled 2N times where N is a number of packages for processing. Maybe compile this once and pass into this function?

Korulag (approved):

PR #318, opened 2022-10-08
Fixed package exporter to use less 'chown' operations

Korulag:


javihernandez (approved):
Looks good, works fine (to the extent I can tell) and doesn't break current workflow. It would be great if we can provide some instructions around the script and the overall exporting process.

anfimovdm (approved):

PR #319, opened 2022-10-10
Fixed NoneType error, reduced amount calls to DB

anfimovdm:


alws/crud/errata.py
```diff
@@ -825,6 +813,22 @@ async def release_errata_packages(
                 "sum_type": "sha256",
             }
         )
+        if rpm_module:
```
javihernandez, line 816: This will always be `None`, isn't it?

anfimovdm: No, we can get `rpm_module` below

PR #326, opened 2022-10-18
Fixed incorrect additional source artifacts in modules metadata

Korulag:


alws/build_planner.py
```diff
@@ -3,6 +3,8 @@
 import typing
 import collections
 import itertools
+import re
+from collections import defaultdict
```
anfimovdm, line 7: Unneeded import

alws/build_planner.py
```diff
@@ -189,6 +190,30 @@ async def prepare_module_index(
...
+                continue
+            for artifact in ref.added_artifacts:
+                parsed_artifact = RpmArtifact.from_str(artifact)
+                module_id = module_id_regex.search(
+                    parsed_artifact.release).groupdict().get('module_id', '')
+                if parsed_artifact.arch in allowed_arches:
+                    src_bin_mapping[module_id].add(parsed_artifact)
+
+        for id_, artifacts in src_bin_mapping.items():
+            if len(artifacts) == 1 and all([i.arch == 'src' for i in artifacts]):
```
anfimovdm, line 213: Can we get rid of `len(...) == 1` here?
also we can throw generator instead of list comprehension
```python3
if all((i.arch == 'src' for i in artifacts)):
    ...
```

anfimovdm (approved):

anfimovdm (approved):

maccelf (approved):

PR #341, opened 2022-10-31
ALBS-724 Allow to set rpm macros in build flavours

maccelf:
Sometimes we need to set some rpm macros only for beta not affecting stable builds.
It is necessary to implement ability to set rpm macros in flavours.
If platform and flavour have the same macros then flavour macros should overwrite platform one i.e. it has higher priority.


alws/crud/platform_flavors.py
```diff
@@ -9,37 +9,61 @@
...
+        .where(models.PlatformFlavour.name == flavour.name)
+        .options(selectinload(models.PlatformFlavour.repos))
+    )
     return db_flavour.scalars().all()
 
 
 async def update_flavour(db, flavour: UpdateFlavour) -> models.PlatformFlavour:
-    db_flavour = await find_flavour_by_id(db, flavour.id)
-    for key in ('name', 'modularity'):
+    db_flavour = await find_flavour_by_name(db, flavour.name)
```
anfimovdm, line 44: `find_flavour_by_name` can return None, please add check for this

alws/routers/build_node.py
```diff
@@ -102,6 +102,15 @@ async def get_task(
                 response['repositories'].append(repo)
     if task.build.platform_flavors:
         for flavour in task.build.platform_flavors:
+            if flavour.data:
+                if flavour.data['mock']['macros']:
```
anfimovdm, line 106: we can fall here with `KeyError` if flavour.data doesn't have `macros` key

scripts/manage_flavours.py
```diff
@@ -0,0 +1,110 @@
...
+    )
+    parser.add_argument(
+        "-R",
+        "--no-remotes",
+        action="store_true",
+        default=False,
+        required=False,
+        help="Disable creation of flavors remotes",
+    )
+    parser.add_argument(
```
anfimovdm, line 32: we don't use this argument

scripts/manage_flavours.py
```diff
@@ -0,0 +1,110 @@
...
+    repository as repo_crud,
+)
+from alws.schemas import platform_flavors_schema, repository_schema
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        "manage_flavours", description="Flavor manage script. Creates flavors"
+    )
+    parser.add_argument(
```
anfimovdm, line 24: same as below

scripts/manage_flavours.py
```diff
@@ -0,0 +1,110 @@
...
+        default=False,
+        required=False,
+        help="Enable verbose output",
+    )
+    return parser.parse_args()
+
+
+async def update_flavour(flavour_data: dict, logger: logging.Logger):
+    async with database.Session() as db:
+        flavour = await pf_crud.find_flavour_by_name(db, flavour_data["name"])
```
anfimovdm, line 68: we get db_flavour twice, here and in `pf_crud.update_flavour`

anfimovdm (changes requested):

anfimovdm (approved):

Korulag (approved):

PR #344, opened 2022-11-01
Added script for releasing errata records through API

anfimovdm:


maccelf (approved):

scripts/release_new_errata_records.py
```diff
@@ -0,0 +1,180 @@
...
+def parse_args():
+    parser = argparse.ArgumentParser(
+        "errata-releaser",
+        description="Releases updateinfo records through ALBS API",
+    )
+    parser.add_argument(
+        "-d",
+        "--date",
+        type=str,
+        default="2022-11-01",
```
soksanichenko, line 31: I think you can use current date

scripts/release_new_errata_records.py
```diff
@@ -0,0 +1,180 @@
...
+        required=False,
+        help="Date for start releasing (default: '2022-11-01')",
```
soksanichenko, line 33: Use [argparse.ArgumentDefaultsHelpFormatter formatter](http://docs.python.org/library/argparse.html#formatter-class) and argparse includes a default value automatically

https://stackoverflow.com/questions/12151306/argparse-way-to-include-default-values-in-help

javihernandez (approved):

scripts/release_new_errata_records.py
anfimovdm: Thanks, fixed

PR #346, opened 2022-11-02
Added script for checking modules in errata records [WIP]

anfimovdm:


anfimovdm:
It's not actual now, because we can't trust our modules in repositories

PR #347, opened 2022-11-02
ALBS-654: Added centos7 config

zklevsha:


reference_data/centos7.yml
```diff
@@ -0,0 +1,116 @@
+---
+- name: CentOS7
+  distr_type: rhel
+  distr_version: '7'
+  #? is this email correct ?#
+  contact_mail: packager@centos.org 
+  copyright: Copyright {{ year }} CentOS
+  test_dist_name: centos7
+  modularity:
```
maccelf, line 9: Centos 7 does not have modules

reference_data/centos7.yml
```diff
@@ -0,0 +1,116 @@
...
+    #? I cant find CentOS alternative for modified_packages.yml ?#
+    modified_packages_url: 'https://git.almalinux.org/almalinux/modified_packages/raw/branch/a8/modified_packages.yml'
+    packages_git: 'https://git.centos.org/projects/rpms/%2A'
+    #? what tag to use here ?#
+    git_tag_prefix:
+      modified: a8
+      non_modified: c8
+    versions:
```
Korulag, line 17: These are module-related versions, they are not needed for CentOS 7. But we need to check if this data is required to store the entry in the database or during the build creation process

reference_data/centos7.yml
```diff
@@ -0,0 +1,116 @@
...
+  
+  arch_list:
+    - x86_64
+    - aarch64
+  copy_priority_arches:
+    - x86_64
+    - aarch64
+  #? I ve removed weak_arch_list since we dont work with those architectures ?# 
+  data:
+    definitions:
```
Korulag, line 59: Check with Andrew on what should be in these definitions please

reference_data/centos7.yml
```diff
@@ -0,0 +1,116 @@
...
+      #? is this teamname/email correct ?#
+      packager: CentOS Packaging Team <packager@centos.org>
+      vendor: CentOS
+    mock:
+      chroot_setup_cmd: 'install bash bzip2 coreutils cloudlinux-release '
+                         'findutils gawk grep gzip info redhat-rpm-config '
+                            'sed shadow-utils tar unzip util-linux-ng which '
+                            'xz yum git'
+      #? I now sure how it will work since CentOS7 uses yum
+      dnf_common_opts:
```
Korulag, line 70: Centos 7 has yum as package manager, take a look at old build system platform configs for correct set of options

reference_data/centos7.yml
```diff
@@ -0,0 +1,116 @@
...
+        "%_pesign": "/usr/local/bin/pesign"
+        "%_pesign_client": "/usr/local/bin/pesign-client"
+        "%with_modsign": "1"
+        "%modsign_os": "centos7"
+    mock_dist: el7
+    timeout: 43200
+    yum:
+      best: false
+      module_platform_id: platform:el7
+  #? There are no repos for aarch64 ?#
```
Korulag, line 97: http://mirror.centos.org/altarch/7/

reference_data/centos7.yml
```diff
@@ -0,0 +1,116 @@
...
+  repositories:
+    - arch: x86_64
+      name: centos-7-baseos
+      type: rpm
+      remote_url: http://mirror.centos.org/centos/7/os/x86_64/
+      export_path: centos/7/os/x86_64/
+      production: true
```
Korulag, line 104: These are not "production" repos by our standards, since we do not export data into them, so `production` should be `False` and `export_path` is not needed

Korulag (changes requested):

reference_data/centos7.yml
zklevsha: @maccelf shold I remove all block modulatiry alltoghether?

maccelf: I think yes

anfimovdm (approved):

javihernandez (approved):

Korulag (approved):

PR #350, opened 2022-11-07
Remove beta specific macros for now

andrewlukoshko:
We're going to enable this later


javihernandez (approved):
LGTM
@andrewlukoshko, I'd say that you'll tell us when the time is come. I've created ALBS-747, just feel free to bring to our attention/create a PR when they need to be re-added

Korulag (approved):

PR #351, opened 2022-11-07
ALBS-702

zklevsha:


scripts/packages_exporter.py
```diff
@@ -42,6 +43,10 @@
 
 KNOWN_SUBKEYS_CONFIG = os.path.abspath(os.path.expanduser(
     '~/config/known_subkeys.json'))
+LOG_DIR = Path.home() / 'exporter_logs'
```
Korulag, line 46: Add check that directory is present on the file system

Korulag (approved):

PR #356, opened 2022-11-09
Added database connection pooling as well as its proper utilization

Korulag:


alws/database.py
```diff
@@ -15,7 +15,11 @@
 
 DATABASE_URL = settings.database_url
 
-engine = create_async_engine(DATABASE_URL, poolclass=NullPool)
+engine = create_async_engine(
+    DATABASE_URL, poolclass=AsyncAdaptedQueuePool,
+    pool_size=10, max_overflow=5, pool_pre_ping=True, pool_recycle=1200,
```
javihernandez, line 20: Out of curiosity, what are the reasons for not using the default values for `pool_size` and `max_overflow`?

alws/database.py
```diff
@@ -15,7 +15,11 @@
...
+    echo_pool=True
```
javihernandez, line 21: Like the idea of setting `echo_pool` :+1:

javihernandez (approved):

anfimovdm (approved):

alws/database.py
Korulag: Previously we've seen that they are not enough. So I prefer a bit bumped numbers

PR #363, opened 2022-11-14
Added ability to read module artifacts from beta

Korulag:
If beta flavor is enabled the system will attempt to get artifacts from related beta distribution to create partial update case.

### DIFF
--- a/alws/build_planner.py
+++ b/alws/build_planner.py
@@ -91,6 +91,16 @@ class BuildPlanner:
         if db_flavors:
             self._platform_flavors = db_flavors
 
+    def is_beta_build(self):
+        if not self._platform_flavors:
+            return False
+        found = False
+        for flavor in self._platform_flavors:
+            if bool(re.search(r'(-beta)$', flavor.name, re.IGNORECASE)):
+                found = True
+                break
+        return found
+
     async def create_build_repo(
                 self,
                 platform: models.Platform,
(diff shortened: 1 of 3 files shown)


anfimovdm (approved):

maccelf (approved):

alws/build_planner.py
```diff
@@ -91,6 +91,16 @@ def load_platform_flavors(self, flavors):
         if db_flavors:
             self._platform_flavors = db_flavors
 
+    def is_beta_build(self):
+        if not self._platform_flavors:
```
soksanichenko, line 95: `return bool(self._platform_flavors)`

alws/build_planner.py
```diff
@@ -91,6 +91,16 @@ def load_platform_flavors(self, flavors):
...
+            return False
+        found = False
+        for flavor in self._platform_flavors:
+            if bool(re.search(r'(-beta)$', flavor.name, re.IGNORECASE)):
+                found = True
+                break
```
soksanichenko, line 101:
```
strings = ['123', '456', '789', 'aaa']
for string in strings:
    if bool(found := re.search('aaa', string)):
        break
        return found
 ```

soksanichenko (changes requested):

alws/build_planner.py
Korulag: I cannot just return at this point because it should return only when no flavors are there

alws/build_planner.py
Korulag: I'm still getting used to this syntax, will do the change

soksanichenko: I agree.

soksanichenko: This comment for conversation above)

soksanichenko (approved):

PR #366, opened 2022-11-15
ALBS-691: Do not fail when yum_exclude comes as an empty array

javihernandez:


Korulag (approved):

alws/schemas/build_node_schema.py
```diff
@@ -56,7 +56,9 @@ def add_mock_options(self, options):
                 self.data['mock'][k] = list(set(self.data['mock'][k]))
             elif k == 'yum_exclude':
                 old_exclude = self.data['yum'].get('exclude', '')
-                self.data['yum']['exclude'] = f'{old_exclude} {" ".join(v)}'
+                full_exclude = f'{old_exclude} {" ".join(v)}'.strip()
+                self.data['yum']['exclude'] = \
+                    full_exclude if full_exclude else None
```
soksanichenko, line 61: `'' or None` returns `None`

javihernandez: And that's what I want :smiley: 
I do know that `''` will evaluate to `None`, and I could have done the following:
`self.data['yum']['exclude'] = f'{old_exclude} {" ".join(v)}'.strip()`
But then, `self.data['yum']['exclude']` will be equal to an empty string, which in terms of checking if None will have exactly the same effect, but still, it's an empty string and to me, I prefer it to become `None`. It's a matter of personal taste, but hey, maybe I'm missing something here.

PR #375, opened 2022-11-28
ALBS-787 Fix mock option in build flavor is not added to build config

maccelf:
Mock option in build flavor is not added to build config
It should be work like it works in secure boot builds

### DIFF
--- a/alws/routers/build_node.py
+++ b/alws/routers/build_node.py
@@ -103,14 +103,17 @@ async def get_task(
     if task.build.platform_flavors:
         for flavour in task.build.platform_flavors:
             if flavour.data:
-                if flavour.data['mock']:
-                    if flavour.data['mock']['macros']:
-                        response['platform'].data['mock']['macros'].update(
-                            flavour.data['mock']['macros']
-                        )
-                if flavour.data['definitions']:
-                    response['platform'].data['definitions'].update(
-                        flavour.data['definitions']
+                for key in ("macros", "secure_boot_macros"):
+                    if  "mock" not in flavour.data or key not in flavour.data["mock"]:
+                        continue
+                    if key not in response["platform"].data["mock"]:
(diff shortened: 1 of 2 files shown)


reference_data/platforms.yaml
```diff
@@ -61,7 +61,7 @@
       environment:
         DFLTCC: '0'
       use_bootstrap_container: false
-      macros:
+      secure_boot_macros:
```
Korulag, line 64: We need the same changes in flavors if they have secure boot macroses

maccelf: Now we don't have secure boot macroses in flavors

alws/routers/build_node.py
```diff
@@ -103,12 +103,22 @@ async def get_task(
...
+                            )
+                        else:
+                            response['platform'].data['mock']['macros'] = flavour.data['mock']['macros']
+                    if 'secure_boot_macros' in flavour.data['mock']:
+                        if 'secure_boot_macros' in response['platform'].data['mock']:
+                            response['platform'].data['mock']['secure_boot_macros'].update(
+                                flavour.data['mock']['secure_boot_macros']
+                            )
+                        else:
+                            response['platform'].data['mock']['secure_boot_macros'] = flavour.data['mock']['secure_boot_macros']
```
anfimovdm, line 120:
```python3
for key in ("macros", "secure_boot_macros"):
    if key not in flavour.data["mock"]:
        continue
    if key not in response["platform"]:
        response["platform"].data["mock"][key] = {}
    response["platform"].data["mock"][key] = flavour.data["mock"][key]
```

anfimovdm (changes requested):

anfimovdm (approved):

Korulag (approved):

PR #395, opened 2023-01-17
ALBS-669 Add a meaningful message error when signing builds

maccelf:
Current we got non-useful message about error Dismiss and 500 error in a browser’s console

### DIFF
--- a/alws/middlewares/__init__.py
+++ b/alws/middlewares/__init__.py
@@ -1,6 +1,8 @@
 from .permissions import handlers as perm_handlers
 from .products import handlers as product_handlers
 from .uploads import handlers as upload_handlers
+from .not_found import handlers as not_found_handler
+from .sign_task import handlers as sign_task_handlers
 
 
 __all__ = ['handlers']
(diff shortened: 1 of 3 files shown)


alws/routers/sign_task.py
```diff
@@ -40,7 +47,28 @@ async def get_sign_tasks(build_id: int = None,
...
+    except ValueError as e:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail=str(e),
+        )
+    except PermissionDenied as e:
+        return HTTPException(
+            status_code=status.HTTP_401_UNAUTHORIZED,
+            detail=str(e),
+        )
```
anfimovdm, line 71: We can define catching exceptions through middleware like we do for products (see `alws/middlewares/products.py`)

anfimovdm (changes requested):

soksanichenko (approved):

anfimovdm (approved):

anfimovdm:
JFYI, you can use constants from `fastapi.status` for HTTP status codes

PR #403, opened 2023-01-27
ALBS-950: Added method to get modules in pulp_client

javihernandez:
Also, updated list_updateinfo_records method so it can be called without specifying the repository

As of context, I made these changes for a script that I'm working on to check for some errors in errata/modules.

### DIFF
--- a/alws/utils/pulp_client.py
+++ b/alws/utils/pulp_client.py
@@ -164,7 +164,7 @@ class PulpClient:
         self,
         name: str,
         stream: str,
-    ) -> typing.Union[typing.List[dict], None]:
+    ) -> typing.Optional[typing.List[dict]]:
         endpoint = 'pulp/api/v3/content/rpm/modulemds/'
         params = {
             'name': name,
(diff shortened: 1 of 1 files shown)


maccelf (approved):

Korulag (approved):

alws/utils/pulp_client.py
```diff
@@ -160,6 +160,21 @@ async def get_rpm_remote(self, name: str) -> typing.Union[dict, None]:
             return None
         return response['results'][0]
 
+    async def get_module_by_name_and_stream(
+        self,
+        name: str,
+        stream: str,
+    ) -> typing.Union[typing.List[dict], None]:
```
anfimovdm, line 167: We can use `typing.Optional` instead of `typing.Union`, because `typing.Optional[...]` is equivalent to `typing.Union[..., None]`

anfimovdm (approved):

alws/utils/pulp_client.py
javihernandez: Done

alws/utils/pulp_client.py
```diff
@@ -756,13 +771,14 @@ async def wait_for_task(self, task_href: str):
     async def list_updateinfo_records(
                 self,
                 id__in: List[str],
-                repository_version: str
+                repository_version: typing.Union[str, None] = None
```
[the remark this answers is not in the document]
javihernandez: Done

anfimovdm (approved):

PR #413, opened 2023-02-08
Use 1 log repository per build

Korulag:
Currently we produce a build log repository for each build task. To make matters worse we produce test log repositories in the same fashion. All this puts a lot of pressure on the Pulp while creating build/making it finished. To avoid these problems we migrate to the flow when we have only 1 build log repository per build and one test log repository.

### DIFF
--- a/alws/crud/build_node.py
+++ b/alws/crud/build_node.py
@@ -424,26 +424,19 @@ async def __process_rpms(db: AsyncSession, pulp_client: PulpClient, task_id: int
 
 
 async def __process_logs(pulp_client: PulpClient, task_id: int,
-                         task_artifacts: list, repositories: list):
-    if not repositories:
+                         task_artifacts: list, repository: models.Repository):
+    if not repository:
         logging.error('Log repository is absent, skipping logs processing')
         return
     logs = []
-    str_task_id = str(task_id)
-    repo = [repo for repo in repositories
-            if repo.name.endswith(str_task_id)]
-    if not repo:
-        logging.error('Log repository is absent, skipping logs processing')
-        return
-    repo = repo[0]
     tasks = [pulp_client.create_entity(artifact)
              for artifact in task_artifacts]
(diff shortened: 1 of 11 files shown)


alws/crud/build_node.py
```diff
@@ -516,8 +462,8 @@ async def __process_build_task_artifacts(
                      if item.type == 'build_log']
     rpm_repositories = [repo for repo in repositories
                         if repo.type == 'rpm']
-    log_repositories = [repo for repo in repositories
-                        if repo.type == 'build_log']
+    log_repository = [repo for repo in repositories
+                      if repo.type == 'build_log'][0]
```
anfimovdm, line 466: we can iterate by `repositories` and `task_artifacts` only once

alws/crud/roles.py
```diff
@@ -26,13 +26,15 @@ async def fix_roles_actions(db: Session, commit: bool = False):
     new_roles = []
 
     for role in roles:
+        print(role.name)
         r_actions = None
         for act_role in RolesList:
+            print(act_role.name)
```
anfimovdm, line 32: debug strings?

alws/crud/test.py
```diff
@@ -16,22 +16,32 @@
...
                 models.BuildTask.status == BuildTaskStatus.COMPLETED
             )
         )).scalars().all()
 
+        build = (await db.execute(select(models.Build).where(
+            models.Build.id == build_id).options(
+            selectinload(models.Build.repos)
+        ))).scalars().first()
+        test_log_repository = [i for i in build.repos
+                               if i.type == 'test_log'][0]
```
anfimovdm, line 36: we can fall here if `build.repos` is empty, we can use here `next` and `generator` instead

anfimovdm (changes requested):

anfimovdm (approved):

PR #414, opened 2023-02-08
Fixed regression of not setting proper info to source RPMs

Korulag:


anfimovdm (approved):
could you please create an unit tests or a ticket for them?

PR #418, opened 2023-02-09
ALBS-297: Added header to nginx conf for decompressing logs, fixed some bugs

anfimovdm:


nginx_configs/albs.conf
```diff
@@ -58,6 +58,10 @@ server {
     }
 
     location /pulp/ {
+        # header for decompressing build/test logs
+        if ($request_uri ~* (\.gz\.log$)) {
```
Korulag, line 62: Usually `.gz` extension is at the end of the file name, could you please adjust (I assume it will be modification on upload part?) ?

soksanichenko (approved):

nginx_configs/albs.conf
```diff
@@ -58,6 +58,10 @@ server {
     }
 
     location /pulp/ {
+        # header for decompressing build/test logs
+        if ($request_uri ~* (\.log$)) {
```
javihernandez, line 62: not .gz?

maccelf (approved):

nginx_configs/albs.conf
anfimovdm: we discussed it with Vasiliy and we decided to left it like it is right now, because .gz at the end means that it would be an archive, I tried to keep archives in our storages, but with this nginx header it will decompress automatically and user will lost ability to open it via file manager

PR #428, opened 2023-02-16
Fix settings path in setup code snippet

m10k:
The code snippet for setting up albs-web-server using docker-compose creates the settings file in an incorrect location, causing an error when resetting the admin password.
The problem is that the volumes directory should not be created in the directory where the docker-compose.yml file is located, but rather the parent directory.

This commit fixes the path in the code snippet so that settings.py is created in the correct location.

This is a very trivial PR, but it took me a bit to understand why the error was shown, so I figured it's worth submitting this. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,12 +58,12 @@ In case you are building containers for the first time, there is how it should b
 
 set -e pipefail
 
-mkdir -p volumes/pulp/settings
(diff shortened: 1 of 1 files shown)


Korulag (approved):

Korulag:
@m10k thanks for your PR, will get this merged soon!

PR #431, opened 2023-02-21
ALBS-952: Ensure that our erratas don't refer to packages coming from the wrong module

javihernandez:
At this point there are two places where this could go wrong:

1. Upon errata creation. We create ErrataToALBSPackages that are present either in production repos or in the build system (because they have been built but not released to production yet). At this point we need to ensure that every mapped package comes from the same module the errata is refering to
2. Upon build done. When build artifacts are processed, we look in the resulting packages that could be proposed for a particular errata package. At this point we need to ensure that the proposed packages are from the same module that the errata refers to


alws/crud/build_node.py
```diff
@@ -380,10 +380,29 @@ def append_errata_package(_, errata_package, artifact, rpm_info):
...
-        for errata_package in errata_packages.scalars().all():
+
+        query = select(models.ErrataPackage).where(
+            sqlalchemy.and_(*conditions)
+        )
+
+        if module_index:
+            # good idea?
+            # we do not build more than one module at a time,
+            # do we?
```
anfimovdm, line 391: I suppose we don't need this comment anymore

alws/utils/pulp_utils.py
```diff
@@ -9,13 +9,83 @@
...
+    RpmModulemdPackages,
     RpmPackage,
 )
 
 
 def get_uuid_from_pulp_href(pulp_href: str) -> uuid.UUID:
     return uuid.UUID(pulp_href.split("/")[-2])
 
+# TODO: Merge with get_rpm_packages_from_repository
+# or keep separate functions but make them share code
```
anfimovdm, line 22: I suppose we can merge them

scripts/albs-952_fill_errata_module.py
```diff
@@ -0,0 +1,30 @@
...
+    async with asynccontextmanager(get_db)() as db, db.begin():
+        errata_records = (await db.execute(
+            select(models.ErrataRecord))).scalars().all()
+        for record in errata_records:
+            match = module_regex.findall(str(record.original_criteria))
+            if match:
+                record.module = match[0]
+                updated_records.append(record)
+        db.add_all(updated_records)
+        await db.commit()
```
anfimovdm, line 26: should we update modules in pulp too or we planning to fix them manually?

alws/utils/pulp_utils.py
```diff
@@ -9,13 +9,83 @@
...
+        )
+        .scalar_subquery()
+    )
+    last_subq = (
+        select(RpmModulemdPackages.package_id)
+        .where(
+            RpmModulemdPackages.modulemd_id.in_(fourth_subq),
+        )
+        .scalar_subquery()
+    )
```
anfimovdm, line 73: are you sure that `RpmModulemdPackages's` exists even if we don't provide artifacts/packages during module creation?

anfimovdm (changes requested):

scripts/albs-952_fill_errata_module.py
javihernandez: Note that the script is meant to update the new column `module` of every errata_record in our db. 
Regarding pulp, as far as I can tell, "should" be fine, look, i.e.:
```
pulp=> select name, module from rpm_updatecollection where update_record_id in (select content_ptr_id from rpm_updaterecord where id='ALSA-2022:7822');
                        name                         |                                                          module                                                          
-----------------------------------------------------+--------------------------------------------------------------------------------------------------------------------------
 almalinux-8-for-i686-appstream-rpms__8_7_default    | {"arch": "i686", "name": "container-tools", "stream": "rhel8", "context": "20125149", "version": 8070020221108190354}
 almalinux-8-for-x86_64-appstream-rpms__8_7_default  | {"arch": "x86_64", "name": "container-tools", "stream": "rhel8", "context": "20125149", "version": 8070020221108190354}
 almalinux-8-for-aarch64-appstream-rpms__8_7_default | {"arch": "aarch64", "name": "container-tools", "stream": "rhel8", "context": "20125149", "version": 8070020221108190354}
 almalinux-8-for-ppc64le-appstream-rpms__8_7_default | {"arch": "ppc64le", "name": "container-tools", "stream": "rhel8", "context": "20125149", "version": 8070020221108190354}
 almalinux-8-for-s390x-appstream-rpms__8_7_default   | {"arch": "s390x", "name": "container-tools", "stream": "rhel8", "context": "20125149", "version": 8070020221108190354}
(5 rows)

```
It would be interesting to run this query in production though.
So yes, let's pair up tomorrow and do some double checking here and there

alws/utils/pulp_utils.py
javihernandez: Right, I've double checked this, and what I observed is that, when processing a build after is finished, we do add artifacts/packages, which results in having 2 modules in pulp:
1. One that is created upon the build creation
2. The other one that is created when the build is finished

i.e. look at [this build](https://build.almalinux.org/build/5781), here's the [link to the modules.yaml](https://build.almalinux.org/pulp/content/builds/AlmaLinux-8-x86_64-5781-br/repodata/87c24e46d7e2adbcca25f2919b924790bde613f5ce60d857faffe1ed5123b00e-modules.yaml) file of that resulting build repo.

alws/crud/build_node.py
javihernandez: done

alws/utils/pulp_utils.py
javihernandez: Not merging them as we will revisit/refactor this in the near future

scripts/albs-952_fill_errata_module.py
javihernandez: For the record, we don't need to fix the modules but the affected updateinfo.xml files. We do know that our pulp modules aren't reliable (as they miss artifacts) and that we should honor the modules.yaml files that are in production.

alws/utils/pulp_utils.py
javihernandez: Finally, we're not taking the modules data from pulp's db but from production modules.yaml files

alws/pulp_models.py
```diff
@@ -377,3 +377,34 @@ def __repr__(self):
     @property
     def pulp_href(self):
         return f"/pulp/api/v3/content/rpm/packages/{str(self.content_ptr_id)}/"
+
```
Korulag, line 380: 2 spaces needed between the classes

alws/pulp_models.py
```diff
@@ -377,3 +377,34 @@ def __repr__(self):
...
+    stream = sqlalchemy.Column(sqlalchemy.Text)
+    version = sqlalchemy.Column(sqlalchemy.Text)
+    context = sqlalchemy.Column(sqlalchemy.Text)
+    arch = sqlalchemy.Column(sqlalchemy.Text)
+
+    @property
+    def nsvca(self):
+        return f"{self.name}:{self.stream}:{self.version}:{self.context}:{self.arch}"
+
+class RpmModulemdPackages(PulpBase):
```
Korulag, line 399: The same

alws/utils/pulp_utils.py
```diff
@@ -12,10 +15,131 @@
     RpmPackage,
 )
 
+from alws.utils.modularity import IndexWrapper
+
 
 def get_uuid_from_pulp_href(pulp_href: str) -> uuid.UUID:
     return uuid.UUID(pulp_href.split("/")[-2])
 
+def get_modules_yaml_from_repo(repo_name: str):
```
Korulag, line 24: 2 spaces needed between functions

Korulag (changes requested):

alws/utils/pulp_utils.py
```diff
@@ -12,10 +15,131 @@
...
+        regex = re.compile(
+            # name-epoch:version-
+            r'^[\w\d\-]+-\d+:[\d\.\w]+-' \
+            # release (this is the capturing group)
+            r'([\d\w\.\-]+\+(?:\w)?\d{1,5}\+(?:\w)?[\d\w]+)(?:\.[\d\w]+)' \
+            r'?(?:\.[\d\w]+)?\.' \
+            # .architecture
+            r'(?:i686|x86_64|src|noarch|aarch64|ppc64le|s390x)?$'
+        )
+        return regex.match(pkg).group(1)
```
anfimovdm, line 100: maybe it could be better to parse NEVRA with `hawkey`?
https://github.com/AlmaLinux/albs-web-server/blob/master/alws/utils/parsing.py#L59

alws/utils/pulp_utils.py
anfimovdm: we can move this function somewhere to `utils/`, I assume it can be helpful in the future

anfimovdm (approved):

alws/crud/build_node.py
```diff
@@ -380,10 +380,26 @@ def append_errata_package(_, errata_package, artifact, rpm_info):
...
-            models.ErrataPackage
-        ).where(sqlalchemy.and_(*conditions)))
-        for errata_package in errata_packages.scalars().all():
+
+        query = select(models.ErrataPackage).where(
+            sqlalchemy.and_(*conditions)
+        )
+
+        if module_index:
+            module = next(module_index.iter_modules())
```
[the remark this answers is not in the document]
javihernandez: Thanks, done!

alws/utils/pulp_utils.py
javihernandez: Right, thanks!

Korulag (approved):

anfimovdm (approved):

javihernandez:
Tests failures aren't related to my changes as they pass on my fork, see https://github.com/javihernandez/albs-web-server/actions/runs/4309149069/jobs/7516194197

PR #440, opened 2023-02-26
Added fastapi-sqla library integration

Korulag:


Korulag:
It may create more problems than resolve, so closing it

PR #452, opened 2023-03-02
ALBS-994 Make "Release Feed" public

maccelf:


palyla (approved):

anfimovdm (changes requested):
please create a test to check that endpoints works without authorization

anfimovdm (changes requested):

anfimovdm (approved):

PR #455, opened 2023-03-06
ALBS-985: Added ability to stop build manually

anfimovdm:


alws/crud/build_node.py
```diff
@@ -189,6 +189,24 @@ async def update_failed_build_items(db: AsyncSession, build_id: int):
         await db.commit()
 
 
+async def mark_build_tasks_as_cancelled(
+    session: AsyncSession,
+    build_id: int,
+):
+    build_tasks = await session.execute(
```
Korulag, line 196: why not straight `update` statement? It will affect only tasks that are idle when the query is executed as well as current code

anfimovdm: I believe that we need to lock this build tasks to prevent from sending them on build nodes and to avoid discrepancies, because we use `with_for_update()` in `get_available_build_task`
https://github.com/AlmaLinux/albs-web-server/blob/master/alws/crud/build_node.py#L33

maccelf (approved):

alws/crud/build_node.py
Korulag: But there we need to do quite a lot of actions before the task will go onto build node hence the lock. Here I see no need for such lock because `update` will happen quickly. Modifying field via loop is much slower than `update` on database side.

anfimovdm: you're right, I forgot that during `DELETE` or `UPDATE` we acquire lock on row
https://www.postgresql.org/docs/current/explicit-locking.html
image

Korulag (approved):

PR #467, opened 2023-03-09
Added timing statistics for releases

Korulag:


alws/crud/release.py
```diff
@@ -137,7 +148,13 @@ async def commit_release(
     )
     product = await __get_product(db, release.product_id)
     releaser = get_releaser_class(product)(db)
-    return await releaser.commit_release(release_id, user_id)
+    release, message = await releaser.commit_release(release_id, user_id)
+    for perf_stat in release.performance_stats:
+        perf_stat.statistics.update(**releaser.stats)
+    db.add(release)
+    await db.commit()
+    release = await releaser.get_final_release(release_id)
```
anfimovdm, line 156: it's unnecessary to return `release` and `message`, because we return in endpoint static message
also please fix annotation for return value type
https://github.com/AlmaLinux/albs-web-server/blob/master/alws/routers/releases.py#L83:#L98

alws/crud/release.py
```diff
@@ -137,7 +148,13 @@ async def commit_release(
     )
     product = await __get_product(db, release.product_id)
     releaser = get_releaser_class(product)(db)
-    return await releaser.commit_release(release_id, user_id)
+    release, message = await releaser.commit_release(release_id, user_id)
+    for perf_stat in release.performance_stats:
+        perf_stat.statistics.update(**releaser.stats)
+    db.add(release)
```
anfimovdm, line 154: is that necessary?

anfimovdm (changes requested):

PR #469, opened 2023-03-09
ALBS-1013 Add more information to errata release logs

maccelf:
Actually, we only write “Successfully released” to it, while I’d say we can add some useful information, such as:

- errata id
- architecture(s)
- is module
- pulp repositories
- packages
- timestamp

### DIFF
--- a/alws/crud/errata.py
+++ b/alws/crud/errata.py
@@ -1200,6 +1200,53 @@ def generate_query_for_release(records_ids: List[str]):
     return query
 
 
+async def get_release_logs(
+    record_id: str,
+    pulp_packages: dict,
+    session: AsyncSession,
+    repo_mapping: dict,
+    db_record: list,
+) -> str:
+    release_log = [
+        f"Record {record_id} successfully released at {datetime.datetime.utcnow()}\n"
+    ]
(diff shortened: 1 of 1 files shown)


alws/crud/errata.py
```diff
@@ -1249,9 +1249,48 @@ async def release_errata_record(record_id: str):
...
+            RpmPackage.version,
+            RpmPackage.release,
+            RpmPackage.arch,
+            RpmPackage.is_modular,
+        ]
+        pulp_pkgs = get_rpm_packages_by_ids(
+            pulp_pkg_ids=[get_uuid_from_pulp_href(pkg) for pkg in pulp_packages],
+            pkg_fields=pkg_fields,
+        )
+        is_modular = False
```
javihernandez, line 1267: Instead of indicating whether the package belongs to a module, I'd prefer to use `db_record.module`, which gives us the `name:stream` of the module. Also, you can get rid of the `is_modular` check below.

javihernandez (commented):
What if `bulk_errata_records_release` endpoint is used? Shouldn't we add the same information to the release logs?

javihernandez (approved):

PR #473, opened 2023-03-13
ALBS-1039: Add ability to show updateinfo.xml in Errata UI

anfimovdm:


alws/crud/errata.py
```diff
@@ -1387,3 +1387,93 @@ async def bulk_errata_records_release(records_ids: List[str]):
...
+    pulp_client = PulpClient(
+        settings.pulp_host,
+        settings.pulp_user,
+        settings.pulp_password,
+    )
+    errata_records = await pulp_client.list_updateinfo_records(
+        id__in=[record_id],
+    )
+    if not errata_records:
+        return f"{record_id} not found in pulp"
```
javihernandez, line 1459: why not returning `None` here and then, returning a 404 on the router side of things? Similar to what we do [here](https://github.com/AlmaLinux/albs-web-server/blob/master/alws/routers/errata.py#L35-L49)

javihernandez (approved):

maccelf (approved):

PR #478, opened 2023-03-19
ALBS-699 Ensure that we placed artifacts according to specification

maccelf:


tests/fixtures/pulp.py
```diff
@@ -164,11 +164,22 @@ async def func(*args, **kwargs):
 @pytest.fixture
 def create_module(monkeypatch):
     async def func(*args, **kwargs):
+        _, modules_yaml, name, stream, release, arch = args
```
javihernandez, line 167: I'm always getting `virt` module for `x86_64`, but in `modules_yaml` I get both `virt` and `virt-devel`. Need to do something about this

tests/fixtures/dramatiq.py
```diff
@@ -78,6 +106,69 @@ async def build_done(
...
+                }
+                for name in get_artifacts(
+                    src_url=build_task.ref.url,
+                    all_artifacts=all_artifacts,
+                    beholder_artifacts=beholder_artifacts,
+                    task_arch=build_task.arch,
+                )
+            ]
+        }
+        await safe_build_done(session, BuildDone(**payload))
```
javihernandez, line 168: Might need to do side-by-side comparison with a real virt module build, but I'd say maybe we should mock the `artifacts` block directly rather than dynamically generating them here.

maccelf:
Too hard to rebase

PR #497, opened 2023-03-28
ALBS-1070: Replace GET request to pulp in releases

anfimovdm:


alws/config.py
```diff
@@ -19,7 +19,7 @@ class Settings(BaseSettings):
     test_task_scheduler_enabled: bool = False
 
     package_beholder_enabled: bool = True
-    beholder_host: typing.Optional[str] = 'http://beholder-web:5000'
```
Korulag, line 22: It can be disabled (theoretically), hence it was optional

Korulag (approved):

PR #530, opened 2023-04-25
Small fixes and code improvements

Korulag:
- Simplified GitHub Oauth client setup since underlying library now has pretty much the same code as we do, so there's no point in subclassing;
- Take CANCELLED build task status into account when finalizing the build;
- Added script to generate long-living JWT tokens for service needs.

### DIFF
--- a/scripts/generate_token.py
+++ b/scripts/generate_token.py
@@ -1,0 +1,58 @@
+import argparse
+import asyncio
+import contextlib
+import sys
+import os
+
+from fastapi_users.authentication.strategy import JWTStrategy
+from sqlalchemy.future import select
+
+sys.path.append(os.path.dirname(os.path.dirname(__file__)))
+
+from alws.dependencies import get_db
+from alws.auth.dependencies import get_user_db
+from alws.auth.user_manager import get_user_manager
+from alws.models import User
+
+
+def parse_args():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-e', '--email', required=False, type=str,
+                        default=None,
+                        help='User e-mail')
+    parser.add_argument('-u', '--username', required=False, type=str,
+                        default=None,
+                        help='User e-mail')
+    parser.add_argument('-s', '--secret', required=True, type=str,
+                        help='JWT secret')
+    return parser.parse_args(sys.argv[1:])
+
+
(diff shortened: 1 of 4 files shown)


javihernandez (approved):

scripts/generate_token.py
```diff
@@ -0,0 +1,58 @@
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-e', '--email', required=False, type=str,
+                        default=None,
+                        help='User e-mail')
+    parser.add_argument('-u', '--username', required=False, type=str,
+                        default=None,
+                        help='User e-mail')
+    parser.add_argument('-s', '--secret', required=True, type=str,
+                        help='JWT secret')
+    return parser.parse_args(sys.argv[1:])
```
anfimovdm, line 28: there is no need to pass args from `sys.argv`, `argparse` module will do it for you
https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.parse_args

scripts/generate_token.py
```diff
@@ -0,0 +1,58 @@
...
+    get_user_db_context = contextlib.asynccontextmanager(get_user_db)
+    get_user_manager_context = contextlib.asynccontextmanager(get_user_manager)
+    async with get_async_session_context() as session:
+        conds = []
+        if email:
+            conds.append(User.email == email)
+        if username:
+            conds.append(User.username == username)
+        if not email and not username:
+            raise ValueError("Need to specify either email or username")
```
anfimovdm, line 43: I'd prefer to check this variables before opening the session, up to you

scripts/generate_token.py
```diff
@@ -0,0 +1,58 @@
...
+    parser.add_argument('-u', '--username', required=False, type=str,
+                        default=None,
+                        help='User e-mail')
+    parser.add_argument('-s', '--secret', required=True, type=str,
+                        help='JWT secret')
+    return parser.parse_args(sys.argv[1:])
+
+
+async def gen_token(secret: str, email: str = None, username: str = None):
+    strategy = JWTStrategy(secret, lifetime_seconds=1 * 31557600)
```
anfimovdm, line 32: maybe we should make `lifetime_seconds` are parameterize?

PR #581, opened 2023-07-05
Fix ValidationError in build creation

anfimovdm:


alws/routers/builds.py
```diff
@@ -35,21 +27,20 @@
...
-    except PermissionDenied:
-        return HTTPException(
-            status_code=status.HTTP_401_UNAUTHORIZED,
-            detail='User has no permissions to create build'
-        )
+    build: build_schema.BuildCreate,
+    user: models.User = Depends(get_current_user),
+    db: AsyncSession = Depends(get_db),
+):
+    return await build_crud.create_build(db, build, user.id)
```
Korulag, line 34: Why `PermissionDenied` handling is removed?

anfimovdm: because we already have middleware that catches `PermissionDenied`
https://github.com/AlmaLinux/albs-web-server/blob/master/alws/middlewares/permissions.py#L14

Korulag (approved):

maccelf (approved):

PR #601, opened 2023-07-25
Added exception handling to bootstrap repositories script

palyla:
image


anfimovdm (approved):
lgtm, btw which repository you weren't able to sync?
maybe it has outdated URL and we should fix it too

javihernandez (approved):

Korulag (approved):

Korulag:
No, it's GCP repos. They are empty thus nothing to sync

PR #603, opened 2023-07-26
AlmaLinux/build-system#57: Mount the sources to a container (albs-web-server, albs-sign-node) instead store them into a container

soksanichenko:
- Copying the code into the images is removed
- Pull the already built image from quay.io

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,8 +7,8 @@ RUN mkdir -p /code && \
                 python3-libmodulemd1 modulemd-tools python-gobject git -y && \
     yum clean all
 RUN curl https://raw.githubusercontent.com/vishnubob/wait-for-it/master/wait-for-it.sh -o wait_for_it.sh && chmod +x wait_for_it.sh
-COPY requirements.txt /tmp/requirements.txt
+COPY requirements.txt /code/requirements.txt
 RUN cd /code && virtualenv -p python3.9 --system-site-packages env && source env/bin/activate \
-    && pip3 install --upgrade pip && pip3 install -r /tmp/requirements.txt --no-cache-dir
+    && pip3 install --upgrade pip && pip3 install -r /code/requirements.txt --no-cache-dir
 WORKDIR /code
-CMD ["/bin/bash", "-c", "source env/bin/activate && uvicorn --workers 4 --host 0.0.0.0 alws.app:app"]
+CMD ["/bin/bash", "-c", "source env/bin/activate && pip3 install --upgrade pip && pip3 install -r requirements.txt --no-cache-dir && uvicorn --workers 4 --host 0.0.0.0 alws.app:app"]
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -7,10 +7,8 @@ RUN mkdir -p /code && \
...
-COPY requirements.txt /tmp/requirements.txt
+COPY requirements.txt /code/requirements.txt
 RUN cd /code && virtualenv -p python3.9 --system-site-packages env && source env/bin/activate \
-    && pip3 install --upgrade pip && pip3 install -r /tmp/requirements.txt --no-cache-dir
-COPY alws /code/alws
-COPY tests /code/tests
+    && pip3 install --upgrade pip && pip3 install -r /code/requirements.txt --no-cache-dir
 WORKDIR /code
-CMD ["/bin/bash", "-c", "source env/bin/activate && uvicorn --workers 4 --host 0.0.0.0 alws.app:app"]
+CMD ["/bin/bash", "-c", "source env/bin/activate && pip3 install --upgrade pip && pip3 install -r requirements.txt --no-cache-dir && uvicorn --workers 4 --host 0.0.0.0 alws.app:app"]
```
anfimovdm, line 14: what a purpose of installing dependencies in entrypoint if they were installed in previous layers?

docker-compose.yml
```diff
@@ -63,16 +63,14 @@ services:
...
     env_file:
       - vars.env
     ports:
       - "8088:8000"
-    build:
-      dockerfile: Dockerfile
-      context: .
     volumes:
       - "./alws:/code/alws"
+      - "./requirements.txt:/code/requirements.txt"
```
anfimovdm, line 73: we already copy `requirements.txt` in Dockerfile's, maybe we need to remove it from here or from Dockerfile's then?

docker-compose.yml
```diff
@@ -97,10 +97,14 @@ services:
       context: .
     volumes:
       - "./alws:/code/alws"
+      - "./tests:/code/tests"
```
anfimovdm, line 100: not sure that task_queue really needs volume with tests

anfimovdm (changes requested):
also I have a question, if someone changes `requirements.txt`, will the docker image be rebuilt or not?

docker-compose.yml
soksanichenko: I just mount that we copy into an image.

docker-compose.yml
soksanichenko: Look at the above comment.

Dockerfile
soksanichenko: We pre-build all images and req.txt is single item that's used inside an image. So, we need reinstall requirements if a developer changes them. That's why I mount req.txt to a container, because it can be changed by compare with the internal version in an image.

anfimovdm (approved):

PR #622, opened 2023-08-24
Fix unpaired double quotation in errata html template

metalefty:
Found this I'm working on: https://github.com/AlmaLinux/build-system/issues/39

See also: https://www.w3.org/International/questions/qa-html-encoding-declarations.en

### DIFF
--- a/alws/utils/templates/errata_alma_page.j2
+++ b/alws/utils/templates/errata_alma_page.j2
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html lang="en">
 <head>
-    <meta content="text/html; charset="UTF-8">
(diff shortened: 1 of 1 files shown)


metalefty:
Existing errata HTMLs are not changed by this change. We need to edit existing errata if necessary but I think we can leave existing HTML as is. This is literally a syntax error but web browsers a intelligent enough to auto-correct this error for rendering.

javihernandez (approved):

javihernandez:
Can you please update this PR with current master?

metalefty:
Rebased onto master and force-pushed.

javihernandez (approved):

anfimovdm (approved):

maccelf (approved):

PR #625, opened 2023-08-30
errata: sort updated packages by arch

metalefty:
Initial fix for this: https://github.com/AlmaLinux/build-system/issues/39

Just sorting by arch at this time until we clarify how to handle checksums. See: https://github.com/AlmaLinux/build-system/issues/98


metalefty:
Replaced by #632

PR #636, opened 2023-09-18
Ensure that we placed artifacts according to specification

anfimovdm:
**Specification**:
- The Main module should have an src.rpm for all the binaries in that module (if ALL binaries got into Devel, then src.rpm goes there);
- There cannot be the same artifacts in the Main module and the Devel module;
- Artifacts that are in the filter sections do not go to the Main module but go to the Devel module;
- src.rpm for the excluded arch are going to the Devel module;
- For multilib modules, artifacts with i686 arch go to x86_64 arch.
  - Multilib artifacts are going into the Main module if they are marked like this: [multilib: [x86_64]](https://git.almalinux.org/modules/llvm-toolset/src/commit/8b28dca34c7e34969f11005f3c9af9b625fa4f5e/SOURCES/modulemd.src.txt#L47) in the components section of module yaml (ruby, llvm-toolset);
  - Other multilib artifacts are going into the Devel module (mariadb, virt, subversion).


anfimovdm:
For some reason, `SLOF-0:20210217-1.module_el8.6.0+2880+7d9e3703.src` is missing in the `virt-devel` build module.
I don't know if that is expected behaviour or not. Maybe we should pass `devel=True` to the `ModuleWrapper.add_rpm_artifact` function when adding artifacts in the `alws.crud.build_node.__process_rpms` function.

tests/fixtures/beholder.py
```diff
@@ -0,0 +1,1937 @@
...
+            return copy.deepcopy(beholder_subversion_response)
+        if (
+            'subversion-1.10.2-5.module_el8.6.0+3347+66c1e1d6.src.rpm'
+            in endpoint
+        ):
+            return copy.deepcopy(beholder_subversion_package_response)
+        if '/module/llvm-toolset/rhel8/x86_64/' in endpoint:
+            return copy.deepcopy(beholder_llvm_response)
+        if '/module/llvm-toolset-devel/rhel8/x86_64/' in endpoint:
+            return copy.deepcopy(beholder_llvm_devel_response)
```
anfimovdm, line 1915: please left only one return statement
```python
result = {}
if 'SLOF-20210217-1.module_el8.6.0+2880+7d9e3703.src.rpm' in endpoint:
    result = beholder_slof_response
...
# also you can return new dict without deepcopy
# return {**result}
return copy.deepcopy(result)
```

tests/fixtures/modularity.py
```diff
@@ -1,3 +1,5 @@
+from asyncio import streams
```
anfimovdm, line 1: unused import

tests/test_api/test_builds.py
```diff
@@ -83,3 +84,150 @@ async def test_build_create_without_permissions(
...
+        get_multilib_packages_from_pulp,
+        enable_beholder,
+        mock_beholder_call,
+        multilib_virt_with_artifacts: str,
+        modules_artifacts: dict,
+        virt_build_payload: dict,
+        virt_modular_build: Build,
+        start_modular_virt_build,
+        virt_build_done,
+        tmp_path,
```
anfimovdm, line 101: you can move common fixtures such as `get_multilib_packages_from_pulp`, `enable_beholder` and etc. in `usefixtures` decorator
```python
@pytest.mark.usefixtures(
    "get_multilib_packages_from_pulp",
    "enable_beholder",
    ...
)
class TestModularBuilds(BaseAsyncTestCase):
    ...
```

anfimovdm:
Also, I think we need to disable the following messages in `pylint`
- `C0114: Missing module docstring (missing-module-docstring)`
- `C0115: Missing class docstring (missing-class-docstring)`
- `C0116: Missing function or method docstring (missing-function-docstring)`
- `R0913: Too many arguments (11/5) (too-many-arguments)`
- `R0914: Too many local variables (19/15) (too-many-locals)`

@Korulag @javihernandez @soksanichenko 
What do you think? Maybe we need to disable something else too?

soksanichenko:
Or we can disable them inline

anfimovdm:
> Or we can disable them inline

we have `pyproject.toml` for that

soksanichenko:
> > Or we can disable them inline
> 
> we have `pyproject.toml` for that

I mean we can disable only for a code/function/etc if we don't want refactoring them, but leave it's enabled for a project in global sense

Korulag (approved):

PR #638, opened 2023-09-20
- Mount sources instead copy them into a Docker image

soksanichenko:
- Function `restart_build_task_tests` is fixed: call `create_test_tasks` contains `db.begin` so we need move out call from another `db.begin`

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -329,9 +329,9 @@ services:
               python tests_cacher.py'"
 
   pulp:
-    image: pulp/pulp
+    # image: pulp/pulp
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -335,9 +329,9 @@ services:
               python tests_cacher.py'"
 
   pulp:
-    # image: pulp/pulp
+    image: pulp/pulp
```
anfimovdm, line 332: don't uncomment this tag

anfimovdm (changes requested):
tests are broken

PR #642, opened 2023-09-22
- Fix GH CI

soksanichenko:
We should remove a product's team in tests because ALBS creates a new team per each new product


Korulag (approved):

anfimovdm (commented):
please don't merge that, I think that some fixture is incorrect and we need fix it
I'll check a bit later

maccelf (approved):

anfimovdm (approved):

PR #656, opened 2023-10-12
Add a Github action for checking alembic migrations

anfimovdm:


anfimovdm:
Here is an example of a job with an incorrect revision order
https://github.com/AlmaLinux/albs-web-server/actions/runs/6497190951/job/17645725105

anfimovdm:
Please don't merge PR before I change `pull_request` on `pull_request_target`
I did that to demonstrate my changes in that PR

.github/workflows/migrations-check.yml
```diff
@@ -0,0 +1,49 @@
...
+      - uses: isbang/compose-action@v1.5.1
+        name: Run docker-compose
+        with:
+          up-flags: "--build --no-deps"
+          down-flags: "--volumes"
+          services: |
+            db
+            web_server
+      - name: Check migratrions
+        run: docker-compose run --rm --no-deps web_server bash -c 'source env/bin/activate && alembic --config alws/alembic.ini upgrade head'
```
Korulag, line 49: This may pass if the situation will be like with Stepan's migration. We need `alembic check` here instead

Korulag (changes requested):

.github/workflows/migrations-check.yml
anfimovdm: `alembic check` doesn't check migrations, this command checks only pending upgrade operations
https://alembic.sqlalchemy.org/en/latest/api/commands.html#alembic.command.check

anfimovdm: also, I added the rebase action to pull changes from master before checking migrations

Korulag (approved):

PR #658, opened 2023-10-16
Fix build-system #115 issue

Korulag:
This commit fixes issues with deleting builds that have references to errata records and other entities inside.
Reference: https://github.com/AlmaLinux/build-system/issues/115


soksanichenko (approved):

tests/test_crud/test_build.py
```diff
@@ -0,0 +1,18 @@
...
+        self,
+        session,
+        base_platform,
+        base_product,
+        create_errata,
+        build_done,
+        build_for_release,
+    ):
+        await session.close()
+        await build.remove_build_job(session, build_for_release.id)
```
anfimovdm, line 18: I think it's better to add a test for the `/{build_id}/remove/` endpoint instead, because we can cover the endpoint with a test too, rather than only that function

anfimovdm (changes requested):
Also, please rebase your branch, I've recently fixed a bug in the `syntax` action

tests/test_crud/test_build.py
Korulag: Fixed

anfimovdm (approved):

PR #663, opened 2023-10-18
Fix creating new sign key

Korulag:
Fixed internal server error when creating new sign key if it already exists or when the platform from the payload is not on the database. https://github.com/AlmaLinux/build-system/issues/66


tests/fixtures/sign_keys.py
```diff
@@ -0,0 +1,49 @@
...
+from alws.models import Platform, SignKey
+from alws.schemas.sign_schema import SignKeyCreate
+
+BASIC_SIGN_KEY_PAYLOAD = {
+    "name": "Test key",
+    "description": "Test sign key",
+    "keyid": "1234567890ABCDEF",
+    "fingerprint": "1234567890ABCDEF1234567890ABCDEF12345678",
+    "public_url": "no_url",
+}
```
anfimovdm, line 17: why do you use constant instead of fixture?

tests/fixtures/sign_keys.py
```diff
@@ -0,0 +1,49 @@
...
+) -> typing.AsyncIterable[SignKey]:
+    sign_key = await __create_sign_key(session, BASIC_SIGN_KEY_PAYLOAD)
+    yield sign_key
+    await session.execute(delete(SignKey))
+    await session.commit()
+
+
+@pytest.mark.anyio
+@pytest.fixture
+async def sign_key_with_platform(
```
anfimovdm, line 42: unused fixture

tests/test_api/test_sign_keys.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+class TestSignKeys(BaseAsyncTestCase):
+    async def test_create_new_key_incorrect_platform(self):
+        payload = BASIC_SIGN_KEY_PAYLOAD.copy()
+        payload["platform_id"] = 9999
+        response = await self.make_request(
+            "post", f"/api/v1/sign-keys/new/", json=payload
+        )
+        print(response.text)
```
anfimovdm, line 12: I think we can remove this print

anfimovdm (changes requested):

anfimovdm (approved):

zklevsha (approved):

PR #664, opened 2023-10-18
Set up a github action to checks that commit messages follow the commit guidelines

maccelf:
Fixes https://github.com/AlmaLinux/build-system/issues/96

### DIFF
--- a/.github/workflows/commit-message.yml
+++ b/.github/workflows/commit-message.yml
@@ -1,0 +1,29 @@
+name: 'Commit Message Check'
+on:
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - reopened
+      - synchronize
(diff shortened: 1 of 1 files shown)


.github/workflows/commit-message.yml
```diff
@@ -0,0 +1,35 @@
+name: 'Commit Message Check'
+on:
+  pull_request:
+    types:
+      - opened
+      - edited
+      - reopened
+      - synchronize
+  pull_request_target:
```
anfimovdm, line 9: After review, we must keep only the `pull_request_target` event, because if we keep both of them, it will trigger additional check

anfimovdm (approved):

Korulag (approved):

PR #666, opened 2023-10-26
AlmaLinux/build-system#87: ALBS can't build multiplatform builds because it puts sRPM to a wrong repo

soksanichenko:
- Select only right rpm repositories by a platform while processing a build task artifacts
- Update built_srpm_url only for the build tasks of one platform
- HTTP-method for route `get_task` is changed from GET to POST, because it sends a request's body (because https://github.com/swagger-api/swagger-ui/issues/8682 & RFC7231)
- Some typing is changed for better autocompleting and linting
- Select and put noarch packages only among the build tasks of one platform
- Create raw_refs per each existing platform, not only first platform from the list


soksanichenko:
Closed, because non-verified commit.

PR #667, opened 2023-10-26
AlmaLinux/build-system#87: ALBS can't build multiplatform builds beca…

soksanichenko:
…use it puts sRPM to a wrong repo

- Select only right rpm repositories by a platform while processing a build task artifacts
- Update built_srpm_url only for the build tasks of one platform
- HTTP-method for route `get_task` is changed from GET to POST, because it sends a request's body (because https://github.com/swagger-api/swagger-ui/issues/8682 & RFC7231)
- Some typing is changed for better autocompleting and linting
- Select and put noarch packages only among the build tasks of one platform
- Create raw_refs per each existing platform, not only first platform from the list

### DIFF
--- a/alws/schemas/build_schema.py
+++ b/alws/schemas/build_schema.py
@@ -307,7 +307,7 @@ async def get_module_data_from_beholder(
 
 def compare_module_data(
     component_name: str,
-    beholder_data: typing.List[dict],
+    beholder_data: tuple[typing.Any],
     tag_name: str,
 ) -> typing.List[dict]:
     pkgs_to_add = []
--- a/alws/schemas/build_schema.py
+++ b/alws/schemas/build_schema.py
@@ -338,12 +338,12 @@ def compare_module_data(
 async def _get_module_ref(
     component_name: str,
     modified_list: list,
-    platform_prefix_list: list,
+    platform_prefix_list: dict,
     module: ModuleWrapper,
     gitea_client: GiteaClient,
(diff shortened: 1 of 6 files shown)


Korulag (approved):

alws/schemas/build_schema.py
```diff
@@ -338,12 +338,12 @@ def compare_module_data(
...
     component_name: str,
     modified_list: list,
-    platform_prefix_list: list,
+    platform_prefix_list: dict,
     module: ModuleWrapper,
     gitea_client: GiteaClient,
     devel_module: typing.Optional[ModuleWrapper],
     platform_packages_git: str,
-    beholder_data: typing.List[dict],
+    beholder_data: tuple[typing.Any],
```
anfimovdm, line 346: I'm not sure, but I think that we might face an error on the pulp host with `tuple` annotation

anfimovdm (approved):

alws/schemas/build_schema.py
soksanichenko: `asyncio.gather` returns `tuple[Any]`
image

anfimovdm: That's not what I meant, I meant that the pulp host has a lower python version than the web server and we might face an error with builtin type annotation, but that's just a guess

soksanichenko: Doesn't matter pulp host's interpreter of python, because that code is executed under py3.9 on a host of web server

PR #673, opened 2023-11-08
AlmaLinux/build-system#113: ALBS can't release/add a multi-platform b…

soksanichenko:
…uild to a product

- Select a proper repo for a package (by arch, type of package (debug or not) and platform ID) while getting/completing a sign task
- Exclude a repo with sign_key while selecting a product's repos
- It's added endpoint to delete only scheduled release plan (for test and following prod purposes)
- Select packages for only one platform while preparing a release plan
- Comments from linter black are fixed
- The tests are fixed


alws/dramatiq/products.py
```diff
@@ -188,8 +193,12 @@ async def set_platform_for_build_repos(
...
     # some-platform-x86_64-some-build-id-debug-br
     repos_per_platform = {
-        f"{task.platform.name}-{repo.arch}-{build.id}-"
-        f"{repo_debug_dict[repo.debug]}": task.platform
+        # fmt: off
+        (
+            f"{task.platform.name}-{repo.arch}-{build.id}"
+            f"-{repo_debug_dict[repo.debug]}"
+        ): task.platform
+        # fmt: on
```
anfimovdm, line 201: how is `black` trying to correct that? as for me, it looks a bit weird to leave such comments as `fmt` in the code

alws/routers/releases.py
```diff
@@ -120,3 +120,45 @@ async def revert_db_release(
...
+                    f'"{release_id}"'
+                ),
+            }
+        else:
+            await db.delete(release)
+            return {
+                'message': (
+                    f'Scheduled release with ID "{release_id}" is removed'
+                ),
+            }
```
anfimovdm, line 164: I'm not sure that we need this endpoint

anfimovdm (changes requested):

alws/routers/releases.py
Korulag: This is more for developer for now, but actually Eduard already asked for ability to delete releases when they are not started yet

PR #694, opened 2023-12-12
Fix COPR repository URL to be arch- and distribution-agnostic

Korulag:
Resolves: https://cloudlinux.atlassian.net/browse/ALBS-903

### DIFF
--- a/alws/utils/copr.py
+++ b/alws/utils/copr.py
@@ -22,8 +22,9 @@ def generate_repo_config(
 ) -> str:
     # we should clean "http" protocol from host url
     clean_host_name = re.sub(r'^(http|https)://', '', settings.pulp_host)
-    repo_url = re.sub('(x86_64|aarch64|ppc64le|i386|i686|s390x)',
-                      '$basearch', repo.url)
+    repo_url = re.sub(
+        '(x86_64|aarch64|ppc64le|i386|i686|s390x)', '$basearch', repo.url
+    )
     repo_url = re.sub('almalinux-(8|9|10)', 'almalinux-$releasever', repo_url)
     config_template = (
         f"[copr:{clean_host_name}:{ownername}:{product_name}]\n"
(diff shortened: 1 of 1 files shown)


alws/utils/copr.py
```diff
@@ -22,10 +22,14 @@ def generate_repo_config(
 ) -> str:
     # we should clean "http" protocol from host url
     clean_host_name = re.sub(r'^(http|https)://', '', settings.pulp_host)
+    repo_url = re.sub(
+        '(x86_64|aarch64|ppc64le|i386|i686|s390x)', '$basearch', repo.url
```
anfimovdm, line 26: that won't work for products that contain arches in their names
```
In [5]: url = 'https://build.almalinux.org/pulp/content/builds/almalinux-8-beta-s390x-AlmaLinux-8-s390x-dr/'

In [6]: re.sub('(x86_64|aarch64|ppc64le|i386|i686|s390x)', '$basearch', url)
Out[6]: 'https://build.almalinux.org/pulp/content/builds/almalinux-8-beta-$basearch-AlmaLinux-8-$basearch-dr/'
```

alws/utils/copr.py
```diff
@@ -22,10 +22,14 @@ def generate_repo_config(
...
+    )
+    repo_url = re.sub('almalinux-(8|9|10)', 'almalinux-$releasever', repo_url)
```
anfimovdm, line 28: same as above
```
In [7]: re.sub('almalinux-(8|9|10)', 'almalinux-$releasever', url)
Out[7]: 'https://build.almalinux.org/pulp/content/builds/almalinux-$releasever-beta-s390x-AlmaLinux-8-s390x-dr/'
```
also I've noticed that we have different suffixes in urls for some products on the production
for example:
https://build.almalinux.org/product/12 - `<product_name>-AlmaLinux-8-...`
https://build.almalinux.org/product/47 - `<product_name>-almalinux-8-...`

anfimovdm (changes requested):

alws/utils/copr.py
Korulag: Valid point, will fix it

alws/utils/copr.py
Korulag: the almalinux-8-beta-s390x was created before COPR was implemented fully, so repos there have different name format

alws/utils/copr.py
Korulag: Fixed

alws/utils/copr.py
```diff
@@ -22,10 +22,23 @@ def generate_repo_config(
...
+            '$basearch-debug-dr',
+            repo.url,
+        )
+    else:
+        repo_url = re.sub(
+            '(x86_64|aarch64|ppc64le|i386|i686|s390x)-dr',
+            '$basearch-dr',
+            repo.url,
+        )
+    repo_url = re.sub('almalinux-(8|9|10)', 'almalinux-$releasever', repo_url)
```
maccelf, line 37: I am not sure this regular expression will work 
[see](https://regex101.com/r/FLpjfT/1)

maccelf (changes requested):

alws/utils/copr.py
Korulag: This is the thing because some of the products were created outside of the community product creation workflow. It will replace things only for community products which is correct behavior

javihernandez (approved):

PR #698, opened 2023-12-21
Sort RSS feed by pubDate

metalefty:
While here, include only recent 500 entries in RSS feed.

Resolves: https://github.com/AlmaLinux/build-system/issues/155

### DIFF
--- a/scripts/packages_exporter.py
+++ b/scripts/packages_exporter.py
@@ -59,7 +59,10 @@ LOG_FILE = LOG_DIR / f"{LOGGER_NAME}_{int(time())}.log"
 def parse_args():
     parser = argparse.ArgumentParser(
         "packages_exporter",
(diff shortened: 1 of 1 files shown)


Korulag (approved):

scripts/packages_exporter.py
```diff
@@ -365,14 +368,17 @@ async def generate_rss(self, platform, modern_cache):
         dist_version = platform.split('-')[-1]
 
         errata_data = modern_cache['data']
+        sorted_errata_data = sorted(
+            erratum_data, key=lambda k: k['updated_date'], reverse=True
```
javihernandez, line 372: s/erratum_data/errata_data ?

metalefty: Yess, it slipped out during squash & resolving conflicts.

javihernandez (approved):

PR #711, opened 2024-02-01
ALBS-1061: Fix "Restart build tests" functionality

maccelf:
After a "Restart build tests" action only affected tests should restart

Resolves: AlmaLinux/build-system#169

### DIFF
--- a/alws/crud/test.py
+++ b/alws/crud/test.py
@@ -10,7 +10,7 @@ from typing import Dict, List, Optional
 from sqlalchemy import insert, update
 from sqlalchemy.ext.asyncio import AsyncSession
 from sqlalchemy.future import select
-from sqlalchemy.orm import selectinload
+from sqlalchemy.orm import joinedload, selectinload
 from sqlalchemy.sql.expression import func
 
 from alws import models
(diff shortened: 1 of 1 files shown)


alws/crud/test.py
```diff
@@ -266,15 +266,33 @@ async def restart_build_tests(db: AsyncSession, build_id: int):
...
+            continue
+
+        last_reviosion = build_task.test_tasks[-1].revision
+        failed = False
+        for test_task in build_task.test_tasks:
+            if test_task.revision != last_reviosion:
+                continue
+
+            if test_task.status == TestTaskStatus.FAILED:
+                failed = True
```
zklevsha, line 289: Maybe we should break out of loop after failed=True? I don`t see why we need to continue

zklevsha (changes requested):

alws/crud/test.py
zklevsha, line 289: Maybe we should break out of loop after failed=True? I don`t see why we need to continue

zklevsha (changes requested):

alws/crud/test.py
```diff
@@ -266,15 +266,33 @@ async def restart_build_tests(db: AsyncSession, build_id: int):
...
+        build_tasks = await db.execute(query)
+        build_tasks = build_tasks.scalars().unique().all()
         test_log_repository = await __get_log_repository(db, build_id)
-    for build_task_id in build_task_ids:
-        await create_test_tasks(db, build_task_id, test_log_repository.id)
+    for build_task in build_tasks:
+        if not build_task.test_tasks:
+            continue
+
+        last_reviosion = build_task.test_tasks[-1].revision
```
javihernandez, line 282: typo `last_reviosion`, it should be `last_revision`

zklevsha (approved):

Korulag (approved):

PR #716, opened 2024-02-08
ALBS can't release/add a multi-platform build to a product

zklevsha:
Fixes https://github.com/AlmaLinux/build-system/issues/133

- Select a proper repo for a package (by arch, type of package (debug or not) and platform ID) while getting/completing a sign task
- Exclude a repo with sign_key while selecting a product's repos
- It's added endpoint to delete only scheduled release plan (for test and following prod purposes)
- Select packages for only one platform while preparing a release plan
- Comments from linter black are fixed
- The tests are fixed

**This is rebased branch**
Original branch AlmaLinux/build-system-133

### DIFF
--- a/alws/build_planner.py
+++ b/alws/build_planner.py
@@ -134,8 +134,6 @@ class BuildPlanner:
             debug=is_debug,
             platform=platform,
         )
-        if repo_type == 'rpm':
-            repo.platform_id = platform.id
         self._build.repos.append(repo)
 
     async def create_log_repo(
--- a/alws/build_planner.py
+++ b/alws/build_planner.py
@@ -134,6 +134,8 @@ class BuildPlanner:
             debug=is_debug,
             platform=platform,
         )
+        if repo_type == 'rpm':
+            repo.platform_id = platform.id
         self._build.repos.append(repo)
 
     async def create_log_repo(
--- a/alws/crud/release.py
+++ b/alws/crud/release.py
@@ -200,7 +200,7 @@ async def remove_release(
         if release is None:
             return {
                 'message': (
-                    'There is no scheduled release plant with ID '
+                    'There is no scheduled release plan with ID '
                     f'"{release_id}"'
                 ),
             }
--- a/alws/crud/release.py
+++ b/alws/crud/release.py
@@ -204,14 +204,13 @@ async def remove_release(
                     f'"{release_id}"'
                 ),
             }
-        else:
-            if not can_perform(release, user, actions.DeleteRelease.name):
-                raise PermissionDenied(
-                    "User does not have permissions to delete this release"
-                )
-            await db.delete(release)
-            return {
-                'message': (
-                    f'Scheduled release with ID "{release_id}" is removed'
-                ),
-            }
+        if not can_perform(release, user, actions.DeleteRelease.name):
+            raise PermissionDenied(
+                "User does not have permissions to delete this release"
+            )
+        await db.delete(release)
+        return {
+            'message': (
+                f'Scheduled release with ID "{release_id}" is removed'
+            ),
+        }
--- a/alws/crud/release.py
+++ b/alws/crud/release.py
@@ -7,7 +7,10 @@ from sqlalchemy.orm import selectinload
 from sqlalchemy.sql.expression import func
 
 from alws import models
-from alws.errors import DataNotFoundError, ProductError
+from alws.constants import ReleaseStatus
+from alws.errors import DataNotFoundError, ProductError, PermissionDenied
+from alws.perms import actions
+from alws.perms.authorization import can_perform
 from alws.release_planner import get_releaser_class
 from alws.schemas import release_schema
 
@@ -174,3 +177,41 @@ async def revert_release(
     product = await __get_product(db, release.product_id)
     releaser = get_releaser_class(product)(db)
     await releaser.revert_release(release_id, user_id)
+
+
+async def remove_release(
+    db: AsyncSession,
+    release_id: int,
+    user: models.User
+):
+    async with db.begin():
+        release = (
+            (
+                await db.execute(
+                    select(models.Release).where(
+                        models.Release.id == release_id,
+                        models.Release.status == ReleaseStatus.SCHEDULED,
+                    )
+                )
+            )
+            .scalars()
+            .first()
+        )
+        if release is None:
+            return {
+                'message': (
+                    'There is no scheduled release plant with ID '
+                    f'"{release_id}"'
+                ),
+            }
+        else:
+            if not can_perform(release, user, actions.DeleteRelease.name):
+                raise PermissionDenied(
+                    "User does not have permissions to delete this release"
+                )
+            await db.delete(release)
+            return {
+                'message': (
+                    f'Scheduled release with ID "{release_id}" is removed'
+                ),
+            }
--- a/alws/crud/release.py
+++ b/alws/crud/release.py
@@ -8,7 +8,7 @@ from sqlalchemy.sql.expression import func
 
 from alws import models
 from alws.constants import ReleaseStatus
-from alws.errors import DataNotFoundError, ProductError, PermissionDenied
+from alws.errors import DataNotFoundError, PermissionDenied, ProductError
 from alws.perms import actions
 from alws.perms.authorization import can_perform
 from alws.release_planner import get_releaser_class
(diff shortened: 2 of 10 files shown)


alws/release_planner.py
```diff
@@ -315,12 +325,21 @@ async def get_pulp_packages(
                     )
                     if key in modules_to_release:
                         continue
+                    for repo in task.build.repos:
+                        logging.info(
+                            'Repo name "%s", Repo type "%s", '
+                            'Repo platfomr_id "%s"',
```
javihernandez, line 331: typo: platfomr

alws/routers/releases.py
```diff
@@ -120,3 +120,45 @@ async def revert_db_release(
...
+        )
+        if release is None:
+            return {
+                'message': (
+                    'There is no scheduled release plant with ID '
+                    f'"{release_id}"'
+                ),
+            }
+        else:
+            await db.delete(release)
```
javihernandez, line 159: Question, are we fine if we don't check if `current_user` has privileges enough to perform this action?

javihernandez (commented):
Looks good on paper. I left a couple of minor comments, happy to approve after comments are fixed/answered

alws/routers/releases.py
Korulag: I'd say we need to check that user has permissions to operate on releases, yes

zklevsha: @javihernandez @Korulag 
I skimmed over other release endpoints and failed to find examples on how we are checking permissions for them.
As far as I see, everything is done via `Dependecny(get_current_user)`

are you suggesting to add some additional checks for this and other enpoints? 
Or am I missing something?

javihernandez: When processing the request, you can get the current user and then add a `can_perform` block like this:
```
if not can_perform(release, user, actions.DeleteRelease.name):
    raise PermissionDenied(
        "User does not have permissions to delete this release"
    )
```
This pattern can be found in many places in our codebase.

Actions are defined in https://github.com/AlmaLinux/albs-web-server/blob/master/alws/perms/actions.py, but I see that there's no `actions.DeleteRelease.name`, so you'd need to create it.

javihernandez: And I'd say a release can be deleted by the owner and superusers only.

alws/routers/releases.py
```diff
@@ -120,3 +120,45 @@ async def revert_db_release(
...
+async def delete_release(
+    release_id: int,
+    db: AsyncSession = Depends(get_db),
+    user: models.User = Depends(get_current_user),
+):
+    """
+    Delete only a scheduled release
+    """
+
+    async with db.begin():
```
javihernandez, line 138: Also, I'd say that we can move this logic into a new function `remove_release` inside `alws/crud/release.py`

alws/routers/releases.py
soksanichenko: Fixed

alws/crud/release.py
```diff
@@ -174,3 +177,41 @@ async def revert_release(
...
+                    )
+                )
+            )
+            .scalars()
+            .first()
+        )
+        if release is None:
+            return {
+                'message': (
+                    'There is no scheduled release plant with ID '
```
javihernandez, line 203: s/plant/plan

alws/dramatiq/products.py
```diff
@@ -165,10 +165,15 @@ async def set_platform_for_products_repos(
...
     # some-username-some-product-some-platform-x86_64-debug-dr
     repos_per_platform = {
-        f"{product.owner.username}-{product.name}-{platform.name.lower()}-"
-        f"{repo.arch}-{repo_debug_dict[repo.debug]}": platform
+        # fmt: off
+        (
+            f"{product.owner.username}-{product.name}-{platform.name.lower()}"
+            f"-{repo.arch}-{repo_debug_dict[repo.debug]}"
+        ): platform
+        # fmt: on
```
anfimovdm, line 173: do we really need to turn off the black formatter there?
maybe we can create a function that would return the needed string?

alws/crud/sign_task.py
```diff
@@ -396,6 +422,14 @@ async def complete_gen_key_task(
     return sign_key
 
 
+def _get_mapping_rpm_per_platform(
+    rpms: list[typing.Union[models.SourceRpm, models.BinaryRpm]],
```
maccelf, line 426: please, use type hints from the `typing` module instead, some type hints are not available for older Python versions
We might fall during exports on Pulp host

alws/build_planner.py
```diff
@@ -134,6 +134,8 @@ async def create_build_repo(
             debug=is_debug,
             platform=platform,
         )
+        if repo_type == 'rpm':
+            repo.platform_id = platform.id
```
maccelf, line 138: What is the point of that if we already passed a platform object ?

maccelf (changes requested):

alws/build_planner.py
soksanichenko: Because the filter doesn't work properly. Looks like a bug. May be it's already fixed in SQLALchemy 2.0

alws/dramatiq/products.py
soksanichenko: I think a separate function is overkill

alws/crud/release.py
```diff
@@ -174,3 +177,41 @@ async def revert_release(
...
+                    f'"{release_id}"'
+                ),
+            }
+        else:
```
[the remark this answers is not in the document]
soksanichenko: Why?

alws/crud/release.py
soksanichenko: Done

alws/dramatiq/products.py
```diff
@@ -223,6 +232,8 @@ async def get_packages_to_blacklist(
     failed_build_tasks = [
         tasks[0][0]
         for tasks in tasks_by_ref.values()
+        # task[0] is its ID
+        # task[1] is status. True if it's completed, False if not
         if not any(task[1] for task in tasks)
```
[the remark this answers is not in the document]
soksanichenko: Done

alws/crud/release.py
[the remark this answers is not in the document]
anfimovdm: I mean, you can do the following:
```python
if release is None:
    return {...}
if not can_perform(release, user, actions.DeleteRelease.name):
    ...
```

alws/dramatiq/products.py
anfimovdm: and what about using the `join` method then?
I think if you use it, there wouldn't be a need to disable the formatter

alws/build_planner.py
anfimovdm: I think that @maccelf meant that you are changing `platform_id` to the same id because the platform object is still the same

alws/crud/release.py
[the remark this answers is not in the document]
soksanichenko: Thank you. I really forgot about that variant

alws/build_planner.py
soksanichenko: as I remember it can be empty

alws/dramatiq/products.py
soksanichenko: I removed turning off

alws/dramatiq/products.py
```diff
@@ -188,8 +191,12 @@ async def set_platform_for_build_repos(
...
     # some-platform-x86_64-some-build-id-debug-br
     repos_per_platform = {
-        f"{task.platform.name}-{repo.arch}-{build.id}-"
-        f"{repo_debug_dict[repo.debug]}": task.platform
+        # fmt: off
+        (
+            f"{task.platform.name}-{repo.arch}-{build.id}"
+            f"-{repo_debug_dict[repo.debug]}"
+        ): task.platform
+        # fmt: on
```
anfimovdm, line 199: I forgot to leave a comment there earlier regarding removing `fmt: off`
could you remove it there too?

alws/build_planner.py
anfimovdm: `repo.platform_id` would be empty until you make a flush or commit, if I'm not mistaken
if you are using `platform_id` in your changes, then I guess it would be better to pass `platform_id` instead of `platform` object during repo creation

alws/dramatiq/products.py
soksanichenko: Done

maccelf (approved):

anfimovdm (approved):

PR #728, opened 2024-02-15
Move issues to the correct statuses

maccelf:
Resolves: AlmaLinux/build-system#70
Resolves: https://github.com/AlmaLinux/build-system/issues/193
Resolves: https://github.com/AlmaLinux/build-system/issues/191
Resolves: https://github.com/AlmaLinux/build-system/issues/185

### DIFF
--- a/alws/config.py
+++ b/alws/config.py
@@ -68,6 +68,12 @@ class Settings(BaseSettings):
     sentry_dsn: Optional[str] = None
     sentry_traces_sample_rate: float = 0.2
 
+    github_integration_enabled: bool = False
+    github_token: Optional[str] = None
+    github_organization_name: str = 'AlmaLinux'
+    github_project_number: Optional[int] = None
+    github_default_repository_name: str = 'updates'
+
     @property
     def codenotary_enabled(self) -> bool:
         return bool(self.immudb_username) and bool(self.immudb_password)
--- a/alws/crud/build.py
+++ b/alws/crud/build.py
@@ -15,13 +15,6 @@ from alws.errors import BuildError, DataNotFoundError, PermissionDenied
 from alws.perms import actions
 from alws.perms.authorization import can_perform
 from alws.schemas import build_schema
-from alws.utils.github_integration_helper import (
-    find_issues_by_repo_name,
-    get_github_client,
-    move_issues,
-    set_build_id_to_issues,
-)
-from alws.utils.parsing import parse_rpm_nevra
 from alws.utils.pulp_client import PulpClient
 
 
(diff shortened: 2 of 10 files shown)


maccelf:
Blocked by https://github.com/AlmaLinux/albs-github-integration/pull/3

alws/routers/errata.py
```diff
@@ -152,6 +158,18 @@ async def release_errata_record(
     db_record.last_release_log = None
     await session.commit()
     release_errata.send(record_id, force)
+    if settings.github_integration_enabled:
```
javihernandez, line 161: What if release fails? Shouldn't be more appropriate to have the dramatiq task updating the status?

alws/config.py
```diff
@@ -68,6 +68,12 @@ class Settings(BaseSettings):
     sentry_dsn: Optional[str] = None
     sentry_traces_sample_rate: float = 0.2
 
+    github_integration_enabled: bool = False
```
javihernandez, line 71: Question, should these vars be included into albs-deploy?

Korulag: Of course

alws/routers/errata.py
maccelf: You right. I'll move this functionality where we set the status for Errata - "Released"

alws/crud/build.py
```diff
@@ -88,6 +95,36 @@ async def create_build(
...
+            if issues:
+                await set_build_id_to_issues(
+                    github_client=github_client,
+                    issues=issues,
+                    build_id=db_build.id,
+                )
+                await move_issues(
+                    github_client=github_client,
+                    issues=issues,
+                    status="Building",
```
javihernandez, line 121: It would be great to turn these statuses into an enum in albs-github-integration, what do you think? It's not urgent, we can create a ticket for that one and, since it's an easy one, anybody (even new contributors) can take it in the future.

alws/crud/build.py
```diff
@@ -88,6 +95,36 @@ async def create_build(
     await db.commit()
     await db.refresh(db_build)
     start_build.send(db_build.id, build.model_dump())
+    if settings.github_integration_enabled:
```
javihernandez, line 98: I'm wondering if we should do the same as with errata, this is, not updating tasks status until the background task is done. In this case, after the build has 'really' started at the very end of the dramatic task

alws/routers/errata.py
javihernandez: Didn't you move this to dramatiq already?

Korulag (approved):

alws/routers/errata.py
maccelf: Yes, I did... I think something broke after rebase

javihernandez (approved):

alws/config.py
javihernandez: created https://github.com/AlmaLinux/build-system/issues/215

PR #729, opened 2024-02-19
Ensure that OVAL data contains released records only

javihernandez:
Fixes: https://github.com/AlmaLinux/build-system/issues/195

### DIFF
--- a/alws/routers/errata.py
+++ b/alws/routers/errata.py
@@ -55,9 +55,10 @@ async def get_errata_record(
 @router.get("/get_oval_xml/", response_model=str)
 async def get_oval_xml(
     platform_name: str,
+    only_released: bool = False,
     db: AsyncSession = Depends(get_db),
 ):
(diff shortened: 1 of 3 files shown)


alws/routers/errata.py
```diff
@@ -55,9 +55,10 @@ async def get_errata_record(
 @router.get("/get_oval_xml/", response_model=str)
 async def get_oval_xml(
     platform_name: str,
+    only_released: bool = False,
```
Korulag, line 58: Won't this basically preserve current behavior? Or we will be relying on the fact that package exporter will set needed flag?

Korulag (approved):

alws/routers/errata.py
javihernandez: Exactly, I don't want to break current behavior as I'm not sure if any other tooling is using this endpoint. And yes, package exporter will set that flag to `True` to ensure that we don't add unreleased erratas into OVAL data.

PR #767, opened 2024-03-28
Adds support for multiple GPG keys

zklevsha:
Fixes https://github.com/AlmaLinux/build-system/issues/205

### DIFF
--- a/alws/crud/errata.py
+++ b/alws/crud/errata.py
@@ -64,6 +64,7 @@ from alws.utils.pulp_utils import (
     get_rpm_packages_from_repository,
     get_uuid_from_pulp_href,
 )
+from alws.utils.oval_add_gpg_keys import add_multiple_gpg_keys_to_oval
 
 try:
     # FIXME: ovallib dependency should stay optional
@@ -152,10 +153,10 @@ async def get_oval_xml(
         )
 
     records = (await db.execute(query)).scalars().all()
-    return errata_records_to_oval(records)
+    return errata_records_to_oval(records, platform_name)
 
 
-def errata_records_to_oval(records: List[models.NewErrataRecord]):
+def errata_records_to_oval(records: List[models.NewErrataRecord], platform_name: str):
     oval = Composer()
     generator = Generator(
         product_name="AlmaLinux OS Errata System",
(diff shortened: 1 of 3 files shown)


alws/utils/oval_add_gpg_keys.py
```diff
@@ -0,0 +1,289 @@
...
+
+import itertools
+import re
+from copy import deepcopy
+from typing import Any, Dict, List, Optional, Tuple
+
+from almalinux.liboval.composer import Composer, Definition, RpminfoState, RpminfoTest
+
+# Map for GPG key ids for different distributions
+GPG = {8: ['51d6647ec21ad6ea', '2ae81e8aced7258b'], 9: ['d36cb86cb86b3716']}
```
javihernandez, line 85: I'd say we no don't need such map nor passing the `distr_version` param around, as you indicated earlier, only AlmaLinux 8 is affected.

alws/crud/errata.py
```diff
@@ -404,6 +401,10 @@ def errata_records_to_oval(records: List[models.NewErrataRecord]):
                 oval.append_object(
                     get_object_cls_by_tag(obj["type"]).from_dict(obj)
                 )
+    # currently only Almalinux8 have multiple GPG keys. So only this platfrom requires
+    # additional oval processing
```
javihernandez, line 405: Please do keep a link to the [original report](https://github.com/AlmaLinux/build-system/issues/205), so anybody reading this code understands why we have to perform this additional processing

javihernandez (commented):
Great job @zklevsha! LGTM
Left a couple of minor comments, also, can you update the commit message according to our [guidelines](https://github.com/AlmaLinux/build-system/wiki/ALBS:-Commit-guidelines#a-tad-more-complex-one)?

alws/utils/oval_add_gpg_keys.py
zklevsha: @javihernandez 
 I added this as support for 9 distr (and all other future distr) in the future, so maybe we should keep it
What do you think?

javihernandez: I know, and kind of makes sense, but the almalinux 8 key change was such an unfortunate event that the core team have put actions in place in order to avoid this situation again in the future. As I said, it's a minor comment aiming to simplify things a bit, I'm fine if you want to keep it, but personally I don't see this happening again in the future

zklevsha: @javihernandez , ok)
I`ve  applied your recommendations

alws/crud/errata.py
zklevsha: @javihernandez added  link to report.
Thanks for pointing out)

Korulag (approved):

Korulag (approved):

PR #785, opened 2024-04-09
Delete pulp db modules using sync session

javihernandez:
Fixes: https://github.com/AlmaLinux/build-system/issues/251


javihernandez:
Seems like async session is working fine in the end, closing this one.

PR #786, opened 2024-04-10
Fix pytest.yml GH workflow

isudak:
The tests should be able to run now.


.github/workflows/pytest.yml
```diff
@@ -49,21 +49,24 @@ jobs:
...
-            web_server
+          context: .
+          load: true
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+      - name: Start services
+        run: docker compose up -d --no-deps web_server db
       - name: Run unit tests (pytest)
-        run: docker-compose run -v /tmp:/tmp --no-deps web_server bash -o pipefail -c '
-          pytest -vv --ignore alws/ --cov-report term-missing:skip-covered
```
javihernandez, line 64: why we no longer have to ignore the `alws` folder?

isudak: Oh, those `pytest` parameters should go into `pyproject.toml` later. Rolled the command line back to what it was.

javihernandez (approved):

PR #792, opened 2024-04-12
Set up test configuration

isudak:
- Put packages for tests into requirements-tests.txt
- Sort packages in requirements.txt
- Make pyproject.toml like in other `albs-*` repos
- Ignore pytest warnings for the Pydantic package
- Create `test_db` and `web_server_tests` services in docker-compose.yml
- Added `tests/test-alembic.ini` and `tests/test-vars.env` for tests

Resolves: AlmaLinux/build-system/issues/225


.github/workflows/migrations-check.yml
```diff
@@ -30,41 +30,23 @@ jobs:
...
-          mkdir -p /home/runner/work/albs-web-server/alts
-          mkdir -p /home/runner/work/albs-web-server/albs-frontend
-          mkdir -p /home/runner/work/albs-web-server/albs-node
-          mkdir -p /home/runner/work/albs-web-server/albs-sign-node
-          mkdir -p /home/runner/work/albs-web-server/alma-tests-cacher
-          touch /home/runner/work/albs-web-server/alma-tests-cacher/vars.env
-          mkdir -p /home/runner/work/albs-web-server/albs-sign-file
-          touch /home/runner/work/albs-web-server/albs-sign-file/.env
-      - name: Add env file
-        run: |
```
isudak, line 42: These vars are moved to the `tests/test-vars.env` file. It's symlinked on line 35.

Dockerfile
```diff
@@ -18,3 +18,13 @@ RUN <<EOT
 EOT
 
 ADD --chmod=755 https://raw.githubusercontent.com/vishnubob/wait-for-it/master/wait-for-it.sh /
+
```
isudak, line 21: Test packages go into a separate image.

docker-compose.yml
```diff
@@ -2,6 +2,24 @@ version: "3.9"
 
 services:
 
+  test_db:
+    image: postgres:13
+    env_file:
+      - vars.env
+    ports:
+      - "5432:5432"
```
metalefty, line 10: Please don't use this port. This is already used by `db` service. This prevents `db` service to start.

isudak: > This prevents `db` service to start.
That's interesting. What `docker compose` command do you use? I'm asking because the `test_db` service shouldn't be started when `web_server` or `db` service starts.

metalefty: See https://github.com/AlmaLinux/build-system/issues/287

PR #105, opened 2022-02-08
albs-89 add column priority to repositories

monkrobot:
Acceptance criteria:
1. Priority flag exists in DB for each repository;
2. Priority flag is passed in task payload (repository section);
3. Mock config honors priority flag for each repository.

PR #855, opened 2024-06-05
RBAC for Test Repositories - Issue #105

amizhen:
Resolves: https://github.com/AlmaLinux/build-system/issues/105
•Teams must now be assigned to Test Repositories upon creation
•Only members of a Test Repo's team can modify the Test Repo
•Done in conjunction with a pull request in frontend: https://github.com/amizhen/albs-frontend/tree/bs-105


alws/alembic.ini
```diff
@@ -50,7 +50,7 @@ version_path_separator = os  # default: use os.pathsep
 # are written from script.py.mako
 # output_encoding = utf-8
 
-sqlalchemy.url = postgresql+psycopg2://postgres:password@db/almalinux-bs
+sqlalchemy.url = postgresql+psycopg2://postgres:password@db/db
```
javihernandez, line 53: I'd say you must keep this untouched

alws/alembic/versions/e952dc755f9d_add_team_to_test_repos.py
```diff
@@ -0,0 +1,164 @@
...
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('test_repository_role_mapping',
+    sa.Column('test_repository_id', sa.Integer(), nullable=False),
+    sa.Column('role_id', sa.Integer(), nullable=False),
+    sa.ForeignKeyConstraint(['role_id'], ['user_roles.id'], name='fk_test_repositories_role_mapping_role_id', ondelete='CASCADE'),
+    sa.ForeignKeyConstraint(['test_repository_id'], ['test_repositories.id'], name='fk_test_repositories_role_mapping_product_id', ondelete='CASCADE'),
+    sa.PrimaryKeyConstraint('test_repository_id', 'role_id')
+    )
+    op.alter_column('build_releases', 'status',
```
javihernandez, line 28: We need to figure out why all these `nullable=False` are getting introduced in the migration. I bet is a side effect from recent migration to fastapi-sqla. @Korulag, any ideas?

alws/crud/test_repository.py
```diff
@@ -139,9 +168,61 @@ async def bulk_create_package_mapping(
     await session.flush()
 
 
+def get_team_role_name(team_name: str, role_name: str):
+    return f'{team_name}_{role_name}'
+
+async def create_test_repository_role_mapping(session: AsyncSession, team_name: str):
+    # print('*MappingDB:0')
```
javihernandez, line 175: get rid of print statements

alws/crud/test_repository.py
```diff
@@ -139,9 +168,61 @@ async def bulk_create_package_mapping(
...
+    required_roles = (Contributor, Observer)
```
javihernandez, line 176: I'd say we should maybe restrict to only team >= Contributor - Given that the Observers don't have the ability to perform any write action

alws/models.py
```diff
@@ -823,6 +823,33 @@ def __repr__(self):
     ),
 )
 
+
+TestRepositoryRoleMapping = sqlalchemy.Table(
```
javihernandez, line 827: Do we really need this?

alws/scripts/albs-gitea-listener/albs-gitea-listener-config.yaml
```diff
@@ -1,12 +1,12 @@
 ---
```
javihernandez, line 1: leave this file untouched

nginx_configs/albs.conf
```diff
@@ -10,9 +10,9 @@ upstream pulp {
     server pulp:80;
```
javihernandez, line 10: same, leave this file untouched

javihernandez (commented):
Thanks @amizhen! Great job on your very first contribution to ALBS!
Overall, looks good, but needs some work. I left some review comments, none of them critical but must be addressed.

javihernandez:
Regarding the tests failure of `tests.test_scripts.test_packages_exporter` ,  it is not related to your changes and it has been fixed in https://github.com/AlmaLinux/albs-web-server/commit/956257df51073db56ad9ec412c88714e85c3c576. All you need to do is to update your code with current master.

alws/crud/test_repository.py
```diff
@@ -139,9 +168,61 @@ async def bulk_create_package_mapping(
...
+    #todo flush if flush see parameters
+    await session.flush()
+
+    return existing_roles
+
+
 async def create_repository(
     session: AsyncSession,
     payload: test_repository_schema.TestRepositoryCreate,
+    # team_id: int,
```
maccelf, line 225: Remove this line

alws/crud/test_repository.py
```diff
@@ -139,9 +168,61 @@ async def bulk_create_package_mapping(
...
 
+def get_team_role_name(team_name: str, role_name: str):
+    return f'{team_name}_{role_name}'
+
+async def create_test_repository_role_mapping(session: AsyncSession, team_name: str):
+    # print('*MappingDB:0')
+    required_roles = (Contributor, Observer)
+    required_actions = (UpdateTest, DeleteTest)
+    role_names = [get_team_role_name(team_name, role.name)
+                       for role in required_roles]
```
maccelf, line 179: Wrong indent

alws/routers/test_repositories.py
```diff
@@ -3,12 +3,14 @@
 from fastapi import APIRouter, Depends, HTTPException, status
 from fastapi_sqla import AsyncSessionDependency
 from sqlalchemy.ext.asyncio import AsyncSession
-
+# from alws.dependencies import get_db
```
maccelf, line 6: Delete this line

alws/crud/test_repository.py
```diff
@@ -139,9 +168,61 @@ async def bulk_create_package_mapping(
...
+                actions_to_add.append(existing_action)
+
+    for role in existing_roles:
+        role_name = f'{team_name}_{role.name}'
+        # print('*MappingDB:', role_name)
+        for action in actions_to_add:
+            if action not in role.actions:
+                # print('*MappingDB:', role_name, action.name)
+                role.actions.append(action)
+            # else:
```
maccelf, line 212: Remove all your unnecessary comments

alws/schemas/test_repository_schema.py
```diff
@@ -45,6 +45,8 @@ class TestRepositoryCreate(BaseModel):
     url: AnyHttpUrlString
     tests_dir: str
     tests_prefix: typing.Optional[str] = None
+    team_id: int
+    owner_id: int
```
maccelf, line 49: Why do we need `owner_id` when we create a test repository? I don't see where we use this variable

maccelf (changes requested):

alws/schemas/test_repository_schema.py
amizhen: You are correct, I removed it and it worked fine

alws/crud/test_repository.py
amizhen: Done

alws/routers/test_repositories.py
amizhen: Done

alws/crud/test_repository.py
amizhen: Done

alws/crud/test_repository.py
amizhen: Done

alws/models.py
amizhen: I think so. Its used here:
`class TestRepository(PermissionsMixin, TeamMixin, Base):
    __tablename__ = "test_repositories"
    id: Mapped[int] = mapped_column(sqlalchemy.Integer, primary_key=True)
    name: Mapped[str] = mapped_column(
        sqlalchemy.Text, nullable=False, unique=True
    )
    url: Mapped[str] = mapped_column(
        sqlalchemy.Text, nullable=False, unique=True
    )
    tests_dir: Mapped[str] = mapped_column(sqlalchemy.Text, nullable=False)
    tests_prefix: Mapped[Optional[str]] = mapped_column(
        sqlalchemy.Text, nullable=True
    )
    packages: Mapped[List["PackageTestRepository"]] = relationship(
        "PackageTestRepository",
        back_populates="test_repository",
        cascade="all, delete",
    )
    team: Mapped["Team"] = relationship("Team", back_populates="test_repositories")
    roles: Mapped[List["UserRole"]] = relationship(
        "UserRole", secondary=TestRepositoryRoleMapping
    )`

alws/crud/test_repository.py
amizhen: Done

alws/alembic/versions/e952dc755f9d_add_team_to_test_repos.py
javihernandez: Ok, this seems to be a side-effect of the recent migration to sqlalchemy 2. I checked one by one in production db and all can be safely updated to set `nullable` to `False`, but some of them need to be handled:
* `build_tasks.mock_options`: these are old builds from 2021 - I'd say that we can update these with an empty `{}` or, if we don't want to, turn this into `nullable=True`
* `builds.mock_options`: these are old builds from 2021-2022 - I'd say that we can update these with a `{'definitions': {}}` or, if we don't want to, make it `nullable=True`
* `builds.cancel_testing ` - These are builds before the introduction of cancel_testing - can set those without cancel_testing with default value (False) 

@Korulag any thoughts? Also, would you like to make these adjustments within this PR? Or do you prefer to create another one?

alws/crud/test_repository.py
```diff
@@ -139,6 +168,44 @@ async def bulk_create_package_mapping(
...
+    existing_roles = (
+        await session.execute(
+            select(UserRole).where(
+                UserRole.name.in_(role_names)
+            ).options(
+                selectinload(UserRole.actions)
+            )
+        )
+    ).scalars().all()
+    # existing_role_names = {r.name for r in existing_roles}
```
javihernandez, line 188: Please delete this and other irrelevant comments

alws/schemas/test_repository_schema.py
javihernandez: @maccelf It's true that we shouldn't really need it, but look at `can_perform` implementation, which  [does the following](https://github.com/AlmaLinux/albs-web-server/blob/e842886770a6302053092d60e5b039b249d41835/alws/perms/authorization.py#L38C5-L38C31). If the object, in this case, the `test_repository` doesn't have an `owner_id`, `can_perform` will fail while trying to access `obj.owner.id`.  Or maybe I'm missing something here.

However, I'd say that we can stop sending it from frontend (is related to [this review comment](https://github.com/AlmaLinux/albs-frontend/pull/475#discussion_r1629717744) of the frontend side of this PR) and manually update the create payload with the current user that is actually performing the action.

alws/scripts/albs-gitea-listener/albs-gitea-listener-config.yaml
javihernandez: up

alws/crud/test_repository.py
amizhen: Done, I checked all of the other files, I believe I've removed them all.

alws/schemas/test_repository_schema.py
javihernandez: Ok, looking at this again, this is what we actually do when creating new products:
* frontend sends user_id -> https://github.com/AlmaLinux/albs-frontend/blob/master/src/pages/CreateProduct.vue#L127
* productCreate schema has owner_id -> https://github.com/AlmaLinux/albs-web-server/blob/master/alws/schemas/product_schema.py#L15
Said that, I'm not sure this is right, maybe we should use the team's owner for new created resources (either products or test_repositories). @maccelf, any thoughts on that?

soksanichenko: > we should use the team's owner for new created resources (either products or test_repositories).

As I remember it's a core idea of RBAC model in ALBS when we was making a prototype.

maccelf: IDK, Is it right if the non-owner of the team creates a test repository and then he will not be an owner of the test repository? Thats weird

javihernandez: Well, that's quite a fair point. In my head, I'd say the resource should belong to the team, as the user is creating it on behalf of the team, but `owner_id` can't be a team, hence my proposal. But yes, let's follow the product's owner_id pattern (the user owns the resource). In that case, we do need this in the payload and this comment can be closed, right?

maccelf: Yes, I think we can keep it as it is

PR #868, opened 2024-06-12
Fix UnboundLocalError in errata endpoint

Korulag:
Build URL was created after the usage in "create_github_issue" function

### DIFF
--- a/alws/utils/github_integration_helper.py
+++ b/alws/utils/github_integration_helper.py
@@ -101,12 +101,12 @@ async def set_build_id_to_issues(
     issues: list,
     build_id: str,
 ):
+    url = urljoin(settings.frontend_baseurl, f"build/{build_id}")
     for issue in issues:
         issue_id = issue["id"]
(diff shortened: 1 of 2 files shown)


alws/utils/github_integration_helper.py
```diff
@@ -101,12 +101,12 @@ async def set_build_id_to_issues(
     issues: list,
     build_id: str,
 ):
+    url = urljoin(settings.frontend_baseurl, f"build/{build_id}")
```
soksanichenko, line 104: Add `+ '/'` for a case when a first argument is not ending to `/`. The concatenation will be wrong in that case.
Same below.

soksanichenko (changes requested):
Please, add a link to task

alws/utils/github_integration_helper.py
Korulag: Fixed by putting validation onto `frontend_baseurl` which adds `/` in the end automatically

PR #872, opened 2024-06-14
Fix nginx service dependencies

pastalian:
Default nginx config (_nginx_configs/albs.conf_) requires sign_file service to be running.

Error log when sign_file service is not running:
```
nginx-1  | nginx: [emerg] host not found in upstream "sign_file:8000" in /etc/nginx/conf.d/albs.conf:14
nginx-1 exited with code 1
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -501,6 +501,7 @@ services:
     depends_on:
       - web_server
       - frontend
+      - sign_file
     logging:
       driver: "json-file"
       options:


pastalian:
@javihernandez It looks like you removed sign_file in 8a43f63. Is there any reason this should not be added?

javihernandez:
Hi @pastalian, thanks for the PR.
IIRC, I removed this dependency because `albs-deploy` is not actually taking care of setting up such a functional `sign_file` service as described [in this issue](https://github.com/AlmaLinux/build-system/issues/204). But I see your point, the nginx conf is expecting to have the `sign_file` service up and running, which is not the case. For this reason, I've created [this task](https://github.com/AlmaLinux/build-system/issues/306) to optionally include the `sign_file` in nginx configuration. So I'd say that we can close this PR. As a workaround right now, you can go ahead and comment the sign_file service definition in nginx.conf and restart the nginx service.

pastalian:
That makes sense. Thank you for clarifying!

PR #924, opened 2024-07-26
Provide the ability to use the "pulp/api"

bklvsky:
- Created `/package_info` endpoint
- Added rate limiting to the endpoint

Resolves: https://cloudlinux.atlassian.net/browse/ALBS-1221

### DIFF
--- a/alws/routers/package_info.py
+++ b/alws/routers/package_info.py
@@ -18,16 +18,13 @@ public_router = APIRouter(
     dependencies=[Depends(RateLimiter(times=5, seconds=1))],
     response_model=List[package_info_shema.PackageInfo],
 )
-async def get_change_log(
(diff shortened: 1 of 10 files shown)


Korulag (approved):

alws/routers/package_info.py
```diff
@@ -0,0 +1,43 @@
...
+        include_fields=[
+            "name",
+            "version",
+            "release",
+            "arch",
+            "changelogs",
+        ],
+        name=package_name,
+    )
+    if release_version:
```
soksanichenko, line 38: Better use obvious condition/behavior/etc.
`release_version is not None`

soksanichenko (changes requested):

PR #929, opened 2024-07-31
Create endpoint to reset package mappings of unreleased erratas

Kwaizer:
Resolves: https://github.com/AlmaLinux/build-system/issues/308

Frontend: https://github.com/AlmaLinux/albs-frontend/pull/504 (**there is a need to discuss the frontend part**)
```
[2024-08-02 14:31:33,515] [PID 36] [Thread-5] [root] [INFO] Packages for records ['4'] have been matched if their date is later than 2024-07-27 16:29:00
2024-08-02 14:35:20 [info     ] engine startup                 engine=Engine(postgresql+psycopg2://postgres:***@pulp/pulp) engine_key=pulp
2024-08-02 14:35:20 [info     ] engine startup                 engine=Engine(postgresql+psycopg2://postgres:***@db/db) engine_key=default
2024-08-02 14:35:20 [info     ] engine startup                 async_engine=<sqlalchemy.ext.asyncio.engine.AsyncEngine object at 0x7faf86bcce00> engine_key=async
2024-08-02 14:35:20 [info     ] engine startup                 async_engine=<sqlalchemy.ext.asyncio.engine.AsyncEngine object at 0x7faf86bd5d80> engine_key=pulp_async
[2024-08-02 14:38:23,290] [PID 36] [Thread-5] [root] [INFO] Packages for records ['5', '4'] have been matched if their date is later than 2024-07-21 16:29:00
```
- added function to find erratas issued after a particular date and check if we have needed records in first place
- added a function to reset packages of these erratas
- added a dramatiq actor to schedule resetting as a background task

### DIFF
--- a/alws/crud/errata.py
+++ b/alws/crud/errata.py
@@ -1717,13 +1717,10 @@ async def reset_matched_errata_packages(record_id: str, session: AsyncSession):
     await session.flush()
 
 
-async def reset_matched_errata_packages_threshold(
-    issued_date_str: str, session: AsyncSession
-):
+async def get_errata_records_threshold(issued_date_str: str, session: AsyncSession):
     issued_date = datetime.datetime.strptime(
         issued_date_str, '%Y-%m-%d %H:%M:%S'
     )
-
     stmt = (
         select(models.NewErrataRecord)
         .where(models.NewErrataRecord.issued_date >= issued_date)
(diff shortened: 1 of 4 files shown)


alws/crud/errata.py
```diff
@@ -1715,3 +1715,66 @@ async def reset_matched_errata_packages(record_id: str, session: AsyncSession):
...
+            selectinload(models.NewErrataRecord.packages).selectinload(
+                models.NewErrataPackage.albs_packages
+            ),
+        )
+        .with_for_update()
+    )
+
+    result = await session.execute(stmt)
+    records = result.scalars().all()
+    print(records)
```
Korulag, line 1747: Remove print statement or make it as log debug entry

alws/crud/errata.py
```diff
@@ -1715,3 +1715,66 @@ async def reset_matched_errata_packages(record_id: str, session: AsyncSession):
...
+    if not records:
+        return f"No unreleased records found after {issued_date_str} including"
+
+    for record in records:
+        items_to_insert = []
+        search_params = prepare_search_params(record)
+        prod_repos_cache = await load_platform_packages(
```
Korulag, line 1754: This is very huge operation, not sure we should keep the whole work in endpoint. Maybe search for entries here and if count > 0 then fire a dramatiq task to actually do the modification while responding in endpoint that work was scheduled?

alws/crud/errata.py
```diff
@@ -1715,3 +1715,66 @@ async def reset_matched_errata_packages(record_id: str, session: AsyncSession):
...
+            selectinload(models.NewErrataRecord.packages).selectinload(
+                models.NewErrataPackage.albs_packages
+            ),
+        )
+        .with_for_update()
+    )
+
+    result = await session.execute(stmt)
+    records = result.scalars().all()
+    await session.flush()
```
anfimovdm, line 1744: I'm not sure that `flush` is necessary there

alws/crud/errata.py
```diff
@@ -1715,3 +1715,66 @@ async def reset_matched_errata_packages(record_id: str, session: AsyncSession):
...
+            )
+            for package in record.packages:
+                matching_packages, _ = await get_matching_albs_packages(
+                    session,
+                    package,
+                    prod_repos_cache,
+                    record.module,
+                )
+                items_to_insert.extend(matching_packages)
+            session.add_all(items_to_insert)
```
anfimovdm, line 1777: You're repeating the same logic as in `reset_matched_errata_packages` function, I think we can move common logic from these functions to a separate function and call it in both of them
Also, you don't have to add to the session new instances for each iteration in that function

alws/crud/errata.py
```diff
@@ -1715,3 +1715,66 @@ async def reset_matched_errata_packages(record_id: str, session: AsyncSession):
...
+            ),
+            selectinload(models.NewErrataRecord.packages).selectinload(
+                models.NewErrataPackage.albs_packages
+            ),
+        )
+        .with_for_update()
+    )
+
+    result = await session.execute(stmt)
+    records = result.scalars().all()
```
anfimovdm, line 1743: You can do the following instead, not critical
```python
records = (await session.execute(stmt)).scalars().all()
```

anfimovdm (changes requested):

alws/crud/errata.py
Kwaizer: Yeah, this can be remade in a separate commit

javihernandez (approved):

anfimovdm (approved):

maccelf (approved):

PR #934, opened 2024-08-05
Update sign certificate for custom product : 36

amizhen:
Resolves: https://github.com/AlmaLinux/build-system/issues/36
• Added archived timestamp and active flag to sign_key table
• Upon new sign_key creation, all keys from the same product get archived
• get_sign_keys endpoint will now only return active sign keys 
(this may have been improper, perhaps it should be filtered in the frontend?)
• Attempts to sign a product with an archived key with throw an error
• sign_key creator can already be seen in PermissionsMixin
• Number of signed packages can be derived in the "build_artifacts" table


alws/alembic/versions/444f9f1f0ac1_.py
```diff
@@ -0,0 +1,30 @@
+"""empty message
```
javihernandez, line 1: Can you please add a message when generating the migration? Right now it's empty and it'd be great to provide a message as we always do. Also, this message is added in the file name too, so it's easily identifiable. I guess you missed passing the  `-m` option when generating the migration.

alws/crud/sign_task.py
```diff
@@ -402,6 +407,15 @@ async def complete_gen_key_task(
...
     )
     roles = [r for r in gen_key_task.product.team.roles if 'signer' in r.name]
+
+    await db.execute(
+        update(models.SignKey).where(
+            and_(models.SignKey.product_id == gen_key_task.product.id,
+                 models.SignKey.active)
+        ).values(active=False, archived=datetime.datetime.utcnow())
+    )
+    await db.commit()
```
anfimovdm, line 417: If I'm not mistaken, calling the `commit` method will close the current session and all actions related to the closed session after that will throw an exception

alws/crud/sign_task.py
```diff
@@ -402,6 +407,15 @@ async def complete_gen_key_task(
         payload.file_name,
     )
     roles = [r for r in gen_key_task.product.team.roles if 'signer' in r.name]
+
+    await db.execute(
+        update(models.SignKey).where(
+            and_(models.SignKey.product_id == gen_key_task.product.id,
+                 models.SignKey.active)
```
anfimovdm, line 414: Don't we need to explicitly compare `models.SignKey.active` to `True`?

alws/models.py
```diff
@@ -1289,6 +1289,12 @@ class SignKey(PermissionsMixin, Base):
     inserted: Mapped[datetime.datetime] = mapped_column(
         sqlalchemy.DateTime, default=datetime.datetime.utcnow()
     )
+    active: Mapped[bool] = mapped_column(
+        sqlalchemy.Boolean, default=True
```
anfimovdm, line 1293: Are you sure that we really need to set it to true by default?

anfimovdm (changes requested):

alws/crud/sign_task.py
amizhen: I don't believe this is the case, I didn't run into such an error when testing. Additionally, the docs don't seem to support this: https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session.commit 
However I may be mistaken here and would defer to others. @javihernandez ?

alws/crud/sign_task.py
amizhen: I don't believe so, models.SignKey.active loaded as a boolean value so you don't need to explicitly check. Additionally, I've tested and verified the behavior, when a key is added only the active keys get archived and the archived one's don't get re-archived. (I verified this by checking the db and the archive times which don't change).

alws/models.py
amizhen: I believe so since a new key is always active.

alws/crud/sign_task.py
anfimovdm: It's fine if it works, probably I mixed it up with something different case or it was a case for sqlalchemy 1.4

alws/crud/sign_task.py
anfimovdm: Okay, I didn't know that it could work in that way

anfimovdm (approved):

javihernandez (approved):

PR #983, opened 2024-09-12
Test integration between packages_exporter and albs-sign-file

bklvsky:
- Enabling test integration between packages_exporter and sign_file
- Setting up albs-sign-file in preflight workflow
- Provisioning testing sign key in the workflow

Resolves: AlmaLinux/build-system#204

### DIFF
--- a/tests/test_scripts/test_packages_exporter.py
+++ b/tests/test_scripts/test_packages_exporter.py
@@ -1,24 +1,21 @@
 import os
 import tempfile
 
-import pytest
-from syncer import sync
-
+from alws.models import SignKey
 from scripts.packages_exporter import Exporter
+from tests.mock_classes import BaseAsyncTestCase
 
 IN_GITHUB_ACTIONS = os.getenv("GITHUB_ACTIONS") is not None
 
 
-@pytest.mark.skip(
-    reason="See https://github.com/AlmaLinux/build-system/issues/204"
-)
-def test_repomd_signer():
-    exporter = Exporter(
-        pulp_client=None, repodata_cache_dir='~/.cache/pulp_exporter'
-    )
-    db_keys = sync(exporter.get_sign_keys())
-    key_id = db_keys[0]['keyid']
-    with tempfile.NamedTemporaryFile() as fp:
-        fp.write(b'Hello world!')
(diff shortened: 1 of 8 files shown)


javihernandez (approved):

tests/test_scripts/test_packages_exporter.py
```diff
@@ -1,24 +1,21 @@
...
+    async def test_repomd_signer(self, sign_key: SignKey):
+        exporter = Exporter(
+            pulp_client=None, repodata_cache_dir='~/.cache/pulp_exporter'
+        )
+        key_id = sign_key.keyid
+        token = await exporter.get_sign_server_token()
+        with tempfile.NamedTemporaryFile() as fp:
+            fp.write(b'Hello world!')
+            res = await exporter.sign_repomd_xml(fp.name, key_id, token)
+            assert res['error'] is None
```
anfimovdm, line 21: I assume that this test will fall locally without setting the needed gpg key and starting the sign-file container, is that correct?

tests/test_scripts/test_packages_exporter.py
```diff
@@ -1,24 +1,21 @@
...
-        res = sync(exporter.sign_repomd_xml(fp.name, key_id))
-        assert res['error'] is None
+class TestPackagesExporter(BaseAsyncTestCase):
+    async def test_repomd_signer(self, sign_key: SignKey):
+        exporter = Exporter(
+            pulp_client=None, repodata_cache_dir='~/.cache/pulp_exporter'
+        )
+        key_id = sign_key.keyid
+        token = await exporter.get_sign_server_token()
+        with tempfile.NamedTemporaryFile() as fp:
```
anfimovdm, line 18: instead of using the `tempfile` module, you can use the built-in `tmp_path` fixture
https://docs.pytest.org/en/stable/how-to/tmp_path.html

anfimovdm:
In my opinion, if we want to check that the sign file service is correctly signing files, it's better to add tests into the corresponding repo instead
in that repo, it will be enough to mock the `sign_repomd_xml` method

tests/test_scripts/test_packages_exporter.py
bklvsky: fixed

anfimovdm (approved):

PR #999, opened 2024-09-25
Unset srpm url if all tasks have failed

Kwaizer:
Resolves: https://github.com/AlmaLinux/build-system/issues/323

### DIFF
--- a/alws/dramatiq/build.py
+++ b/alws/dramatiq/build.py
@@ -157,6 +157,17 @@ async def _start_build(build_id: int, build_request: build_schema.BuildCreate):
 
 async def _build_done(request: build_node_schema.BuildDone):
     async with open_async_session(key=get_async_db_key()) as db:
+        build_task = (
+            (
+                await db.execute(
+                    select(models.BuildTask).where(
+                        models.BuildTask.id == request.task_id
+                    )
+                )
+            )
+            .scalars()
+            .first()
+        )
         try:
             await build_node_crud.safe_build_done(db, request)
         except Exception as e:
@@ -166,22 +177,13 @@ async def _build_done(request: build_node_schema.BuildDone):
                 request.task_id,
                 str(e),
             )
-            build_task = (
-                (
-                    await db.execute(
-                        select(models.BuildTask).where(
-                            models.BuildTask.id == request.task_id
-                        )
-                    )
-                )
-                .scalars()
-                .first()
-            )
             build_task.ts = datetime.datetime.utcnow()
             build_task.error = str(e)
             build_task.status = BuildTaskStatus.FAILED
             await build_node_crud.fast_fail_other_tasks_by_ref(db, build_task)
             await db.flush()
+
+        build_id = build_task.id
 
         # We don't want to create the test tasks until all build tasks
         # of the same build_id are completed.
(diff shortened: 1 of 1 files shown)


Korulag (approved):

alws/crud/build_node.py
```diff
@@ -896,8 +896,9 @@ async def __update_built_srpm_url(
     # if SRPM built we need to download them
     # from pulp repos in next tasks
     if srpm_artifact and build_task.built_srpm_url is None:
+        platform_name = (await build_task.awaitable_attrs.platform).name
```
anfimovdm, line 899: Why do we need it?
Also, I think that's a bad idea to access attributes via `awaitable_attrs`(see https://docs.sqlalchemy.org/en/20/orm/extensions/asyncio.html#sqlalchemy.ext.asyncio.AsyncAttrs and https://docs.sqlalchemy.org/en/20/orm/extensions/asyncio.html#sqlalchemy.ext.asyncio.AsyncSession.run_sync)

alws/dramatiq/build.py
```diff
@@ -166,23 +177,36 @@ async def _build_done(request: build_node_schema.BuildDone):
...
-                .scalars()
-                .first()
-            )
             build_task.ts = datetime.datetime.utcnow()
             build_task.error = str(e)
             build_task.status = BuildTaskStatus.FAILED
             await build_node_crud.fast_fail_other_tasks_by_ref(db, build_task)
             await db.flush()
 
+        build_id = await _get_build_id(db, request.task_id)
```
anfimovdm, line 186: You can get `build_id` from previously loaded `build_task`

alws/dramatiq/build.py
```diff
@@ -166,23 +177,36 @@ async def _build_done(request: build_node_schema.BuildDone):
...
+                update(models.BuildTask).where(
+                    models.BuildTask.build_id == build_id,
+                    models.BuildTask.ref_id == build_task.ref_id,
+                ).values(
+                    status=BuildTaskStatus.FAILED,
+                    built_srpm_url=None,
+                )
+            )
+
+        await db.flush()
```
anfimovdm, line 208: Can't we check it only once when all build tasks are completed?

anfimovdm (changes requested):

alws/crud/build_node.py
Kwaizer: Now I'm not really sure why it was there too :/ 
BTW, you think that accessing values via `awaitable_attrs` is a bad practice or it wouldn't be right in this specific context?

soksanichenko (approved):

alws/crud/build_node.py
anfimovdm: That's not a bad practice, but we should avoid such things in async apps

alws/dramatiq/build.py
```diff
@@ -166,23 +177,23 @@ async def _build_done(request: build_node_schema.BuildDone):
...
+        build_id = build_task.id
+
+        # We need to unset srpm_url if all tasks failed
+        build_tasks = (await db.execute(
+            select(models.BuildTask).where(
+                models.BuildTask.build_id == build_id,
+                models.BuildTask.ref_id == build_task.ref_id,
+                models.BuildTask.arch != 'src',
+            )
+        )).scalars().all()
```
anfimovdm, line 195: Please move that query to the if statement below, because we need it only there

alws/dramatiq/build.py
```diff
@@ -192,7 +203,18 @@ async def _build_done(request: build_node_schema.BuildDone):
         )
 
         if all_build_tasks_completed:
```
anfimovdm, line 205: We can reduce nesting in the following way:
```python
if not all_build_tasks_completed:
     return
failed_tasks = [...]
...
```

alws/dramatiq/build.py
```diff
@@ -192,7 +203,18 @@ async def _build_done(request: build_node_schema.BuildDone):
...
-            build_id = await _get_build_id(db, request.task_id)
+            failed_tasks = [task for task in build_tasks if task.status == BuildTaskStatus.FAILED]
+            if len(failed_tasks) == len(build_tasks):
+                await db.execute(
```
anfimovdm, line 208: Not sure that we need to update these fields via query because we already have build tasks from DB, we can set `build_srpm_url=None` using object attributes
```python
for task in build_tasks:
    task.build_srpm_url = None
```

anfimovdm (changes requested):

alws/dramatiq/build.py
```diff
@@ -209,13 +231,14 @@ async def _build_done(request: build_node_schema.BuildDone):
...
                         str(e),
                     )
-            build_id = await _get_build_id(db, request.task_id)
             await db.execute(
                 update(models.Build)
                 .where(models.Build.id == build_id)
                 .values(finished_at=datetime.datetime.utcnow())
             )
 
+            await db.flush()
```
anfimovdm, line 240: Unnecessary `db.flush`

anfimovdm (changes requested):

anfimovdm (approved):

PR #1006, opened 2024-10-01
Add products exporter

anfimovdm:
Resolves: AlmaLinux/build-system#354

### DIFF
--- a/scripts/add_export_paths_to_product_repos.py
+++ b/scripts/add_export_paths_to_product_repos.py
@@ -19,7 +19,10 @@ def main():
             session.execute(
                 select(Product)
                 .where(Product.is_community.is_(True))
-                .options(joinedload(Product.repositories), joinedload(Product.platforms))
+                .options(
+                    joinedload(Product.repositories),
+                    joinedload(Product.platforms),
+                )
             )
             .scalars()
             .unique()
@@ -27,7 +30,9 @@ def main():
         ):
             if not product.repositories:
                 continue
-            platform_names = '|'.join((platform.name for platform in product.platforms))
+            platform_names = '|'.join(
+                (platform.name for platform in product.platforms)
+            )
             platform_pattern = re.compile(
                 rf'-({platform_names})-(\w+)(-debug|)-dr$',
                 flags=re.IGNORECASE,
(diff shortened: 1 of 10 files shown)


scripts/add_export_paths_to_product_repos.py
```diff
@@ -0,0 +1,48 @@
...
+                .where(Product.is_community.is_(True))
+                .options(joinedload(Product.repositories), joinedload(Product.platforms))
+            )
+            .scalars()
+            .unique()
+            .all()
+        ):
+            if not product.repositories:
+                continue
+            platform_names = '|'.join((platform.name for platform in product.platforms))
```
soksanichenko: Double brackets around the generator are redundant. You can use single brackets.
For example:

```
>>> ', '.join(str(i) for i in range(10))
'0, 1, 2, 3, 4, 5, 6, 7, 8, 9'
>>>
```

scripts/add_export_paths_to_product_repos.py
```diff
@@ -0,0 +1,48 @@
...
+            platform_pattern = re.compile(
+                rf'-({platform_names})-(\w+)(-debug|)-dr$',
+                flags=re.IGNORECASE,
+            )
+            for repo in product.repositories:
+                if repo.arch == 'sign_key':
```
soksanichenko, line 41: I think it's better to use `repo.type == 'sign_key'`

scripts/add_export_paths_to_product_repos.py
```diff
@@ -0,0 +1,48 @@
...
+                    continue
+                regex_result = platform_pattern.search(repo.name)
```
soksanichenko, line 43: Why can't we use `repo.platform.name`?

soksanichenko (changes requested):

scripts/add_export_paths_to_product_repos.py
anfimovdm: It doesn't matter, because when we create a sign key repo we set 'sign_key' for arch and type as well

scripts/add_export_paths_to_product_repos.py
anfimovdm: Because `platform` is missing for products that don't have any builds in it

scripts/add_export_paths_to_product_repos.py
anfimovdm: Fixed

scripts/add_export_paths_to_product_repos.py
soksanichenko: It looks better

maccelf (approved):

PR #1008, opened 2024-10-04
Add ability to add new platforms to the product

maccelf:
Resolves: AlmaLinux/build-system#358

### DIFF
--- a/alws/crud/products.py
+++ b/alws/crud/products.py
@@ -18,7 +18,7 @@ from alws.errors import DataNotFoundError, PermissionDenied, ProductError
 from alws.models import Build, Product, Repository, Team, UserRole
 from alws.perms import actions
 from alws.perms.authorization import can_perform
-from alws.schemas.product_schema import ProductCreate
+from alws.schemas.product_schema import ProductCreate, Platform
 from alws.schemas.team_schema import TeamCreate
 from alws.utils.copr import create_product_repo, create_product_sign_key_repo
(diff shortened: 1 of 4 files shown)


alws/crud/products.py
```diff
@@ -338,18 +364,95 @@ async def modify_product(
...
+            name=repo_name,
+            url=repo_url,
+            arch=arch,
+            pulp_href=pulp_href,
+            type=arch,
+            debug=is_debug,
+            production=True,
+        )
+        db_product.repositories.append(repo)
+        items_to_insert.append(repo)
```
anfimovdm, line 431: Most logic of the new function repeats the same logic as in the `create_product` function, please move common logic to a separate function and reuse it

anfimovdm (changes requested):

anfimovdm (approved):

alws/crud/products.py
```diff
@@ -34,6 +34,63 @@
 logger = logging.getLogger(__name__)
 
 
+async def add_repos_to_product(
+    pulp_client: PulpClient,
+    product: models.Product,
+    owner: models.User,
+    platforms: List[models.Platform]
+):
```
soksanichenko, line 42: add typing for a return value

soksanichenko (commented):
LGTM, except a tiny comment. It's FYI.

anfimovdm (approved):

PR #1035, opened 2024-10-23
- [fixed] getting distribution name of a platform if it contains dashes

soksanichenko:


Korulag (approved):

alws/utils/parsing.py
```diff
@@ -42,9 +42,9 @@ def clean_release(release: str) -> str:
 
 def get_clean_distr_name(distr_name: str) -> str:
     clean_distr_name = re.search(
-        r'^(?P<dist_name>[a-z]+)', distr_name, re.IGNORECASE,
+        r'^(?P<dist_name>[^\d]*)', distr_name, re.IGNORECASE,
     ).groupdict().get('dist_name', '')
-    return clean_distr_name
+    return clean_distr_name.strip('-')
```
anfimovdm, line 47: Instead of additional stripping we can just adjust the regex
```
^(?P<dist_name>\w+(?:-\w+)?)
or
^(?P<dist_name>[a-z]+(?:-[a-z]+)?)
```

soksanichenko: Both doesn't work if a platform name contains more than two dashes.

soksanichenko: Yep, the second option is better, thank you!

PR #1037, opened 2024-10-23
Add new parameter to platform

Kwaizer:
Resolves: https://github.com/AlmaLinux/build-system/issues/368


maccelf (changes requested):
I believe you need to do a migration
Also, make sure that the bootstrap script adds new field to platforms

Kwaizer:
> I believe you need to do a migration Also, make sure that the bootstrap script adds new field to platforms
Forgot to add, thanks for pointing out

soksanichenko (changes requested):
I agree with comment from Maxim. It's the important thing.

maccelf (approved):

Korulag (approved):

javihernandez (approved):

soksanichenko (approved):

Korulag (approved):

PR #1043, opened 2024-11-07
Exclude builds from node

bklvsky:
Implement mechanism for excluding projects on a node

- Change build_node/get_task Request model
- Look up available tasks based on excluded_packages
- black formatting

Resolves:  AlmaLinux/build-system#365

### DIFF
--- a/alws/crud/build_node.py
+++ b/alws/crud/build_node.py
@@ -36,7 +36,7 @@ from alws.utils.modularity import IndexWrapper, RpmArtifact
 from alws.utils.multilib import MultilibProcessor
 from alws.utils.noarch import save_noarch_packages
 from alws.utils.parsing import clean_release, parse_rpm_nevra
-from alws.utils.pulp_client import PulpClient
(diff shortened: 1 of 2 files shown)


alws/crud/build_node.py
```diff
@@ -977,7 +1007,9 @@ async def safe_build_done(
     request: build_node_schema.BuildDone,
 ):
     success = True
-    pulp = PulpClient(settings.pulp_host, settings.pulp_user, settings.pulp_password)
+    pulp = PulpClient(
```
anfimovdm, line 1010: Please replace it with `alws.utils.pulp_client.get_pulp_client`

alws/crud/build_node.py
```diff
@@ -59,6 +72,7 @@ async def get_available_build_task(
                     models.BuildTask.ts < ts_expired,
                     models.BuildTask.ts.is_(None),
                 ),
+                exclude_condition,
```
anfimovdm, line 75: Does the query work if we got `None` here?

anfimovdm (approved):

PR #1084, opened 2024-12-04
Fix cyclic import

Kwaizer:
Resolves: https://github.com/AlmaLinux/build-system/issues/403
- creating `tasks` subdir and moving `event_loop` initialization there so we wouldn't import it from `dramatiq/__init__.py`
- refactoring imports in `routers` and `dramatiq` dirs


Kwaizer:
@anfimovdm how did you checked this?

PR #1099, opened 2025-01-11
Remove codenotary repo from task-queue image

pastalian:
Codenotary repo currently returns 403, causing image builds to fail. Remove this as it is no longer needed since
97a503e14e655108d25af7608cd795bf24186481.

Fixes: https://github.com/AlmaLinux/build-system/issues/412


javihernandez (approved):

javihernandez:
Yes, we removed the installation of `cas` but the repo was left behind, thanks a lot @pastalian! It is now merged :)

PR #1100, opened 2025-01-13
ELSIF-69: Add tests for oval generation

gost-serb:
Resolves: https://github.com/AlmaLinux/build-system/issues/413


soksanichenko (approved):

tests/export_new_errata_rec.py
```diff
@@ -0,0 +1,67 @@
...
+
+if __name__ == '__main__':
+    sqlalchemy_url = os.getenv('SQLALCHEMY_URL')
+    RECORDS_QTY = 20
+
+    engine = create_engine(sqlalchemy_url, echo=False)
+    session = Session(engine)
+    stmt = (
+        select(NewErrataRecord)
+        .where(NewErrataRecord.id.like('ALSA%'))
```
javihernandez, line 58: There might be ALEAs and ALBAs here

tests/export_new_errata_rec.py
```diff
@@ -0,0 +1,67 @@
...
+        .where(NewErrataRecord.criteria != 'null')
+    )
+    errata_records = session.execute(stmt).scalars().fetchmany(RECORDS_QTY)
```
javihernandez, line 61: why not adding a `limit` on the query and saving the db from returning everything, if at the end we're getting only 20?

javihernandez (commented):
Hey @gost-serb, thank you for your contribution!
I left a couple of minor comments, but I also have some other reservations:

1. I don't like the idea of adding 6Mb of samples data for testing (and encouraging people to update them), this could probably make the size of the repo to grow quickly. Can't we just compile sample data only with some key erratas? Like, dunno:
* One simple with one package, i.e.: thunderbird
* One that involves several packages
* One that involves a module
* Another one that involves a devel module (for this one I think you'd need to mock the response from beholder IIRC)

Instead of having 20 random erratas, I'd prefer to only exercise those that bring value to testing such process.

2. Related to my reasoning about the first bullet, I'd say that I don't like the idea of including the `export_new_errata_rec.py` in the tests folder, as it doesn't really has anything to do with testing. For this reason, I'd include it in the scripts folder instead

3. I'd say that serialize/deserialize model functions should live together in some kind of utility file, rather than being spread in two different places

gost-serb:
@javihernandez thanks for the feedback! I'll update PR according to the recommendataions.

scripts/serialize_new_errata_rec.py
```diff
@@ -0,0 +1,139 @@
...
+    logging.basicConfig(level=logging.INFO, handlers=[logging.StreamHandler()])
+
+    sqlalchemy_url = os.getenv('SQLALCHEMY_URL')
+
+    engine = create_engine(sqlalchemy_url, echo=False)
+    session = Session(engine)
+    stmt = select(NewErrataRecord).where(
+        NewErrataRecord.id.in_((
+            'ALSA-2022:4991',
+            'ALSA-2022:4887',
```
javihernandez, line 123: This one is not released and won't be processed (in fact, is not present in the xml sample you provided), so it doesn't really bring any value here.

I'd recommend to use modern and released ones, here's a list:
* ALSA-2025:0281 (single pkg)
* ALSA-2025:0325 (multiple pkgs | including noarch pkgs)
* ALSA-2024:9644 (module | multiple pkgs)
* ALSA-2024:6964 (module | devel module | multiple pkgs)

javihernandez (commented):
@gost-serb I left a small comment regarding the erratas used in testing. Other than that, I'm fine with the implementation and it will be ready to merge after you make the last changes I requested.

Also, since the PR is already including the old large test datasets in the history, we should squash when merging and, probably, remove the references to the ticket(s) from the commit title as they don't really make sense to be there.

javihernandez (changes requested):
Hey @gost-serb, first, thanks for updating it so fast.

I have a last request regarding the json data file. Mostly, I have 2 requests:
* I see that the erratas are duplicated in both json and xml files, can you please get rid of duped ones?
* I also see that we can shrink the size of the json file by removing some fields that we don't need for this test. These are:
  * all `original_*` fields, except the `original_description` one
  * the `last_release_log` field

scripts/serialize_new_errata_rec.py
```diff
@@ -20,9 +20,12 @@
 from alws.models import NewErrataRecord, NewErrataPackage, Platform
 
 
-def serialize_model(instance, seen=None, exclude_models=None):
+def serialize_model(
+    instance, seen=None, exclude_models=None, exclude_columns=None
+):
     seen = set() if seen is None else seen
     exclude_models = [] if exclude_models is None else exclude_models
+    exclude_columns = [] if exclude_columns is None else exclude_columns
```
javihernandez, line 28: I'm curious, is there a reason to don't just set the default values in the function definition? This is:
```
def serialize_model(
    instance, seen=set(), exclude_models=[], exclude_columns=[]
):
```

gost-serb: Technically, in this case it's possible to make them default values since we only read these arguments.
But I think it's safer to keep existing implementation because of how python handles mutable default values.

javihernandez: Ok, I was asking because we don't modify/write the lists or return any of these values. In that case I'd provide the defaults, but I leave it up to you. Thanks for the clarification

gost-serb: Agree but from my experience it's too easy to introduce some changes that modify such argument and forget to move mutable part out from default value.

javihernandez (approved):

anfimovdm (approved):

PR #1154, opened 2025-06-02
Add mock bootstrap images in platform definitions

### LINKED ISSUE #483: Use bootstrap images when possible
After the problems we have experienced when trying to build AlmaLinux 9 kmods in AlmaLinux 10 with recent versions of mock, we have decided to start using bootstrapped chroots when building with mock.

Also, in order to speed up the build time a bit, we have decided to start using bootstrap images to the extent possible. The only exception is x86_64_v2, for which we can not provide an image for mock to use, we'll create the bootstrap from scratch.

javihernandez:
Resolves: https://github.com/AlmaLinux/build-system/issues/483

### DIFF
--- a/reference_data/platforms.yaml
+++ b/reference_data/platforms.yaml
@@ -47,7 +47,7 @@
       releasever: '8'
       environment:
         DFLTCC: '0'
-      use_bootstrap_container: false
+      bootstrap_image: 'quay.io/almalinuxorg/almalinux:8'
       nspawn_args:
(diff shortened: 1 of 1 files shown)


anfimovdm (approved):

andrewlukoshko:
@javihernandez we also need to add
```
config_opts['bootstrap_image_ready'] = True
```
to `mock.cfg` to avoid unnecessary dnf run with metadata download

PR #1174, opened 2025-08-27
Implement a script for build notarization

anfimovdm:
Resolves: https://github.com/AlmaLinux/build-system/issues/504

### DIFF
--- a/scripts/notarize_build.py
+++ b/scripts/notarize_build.py
@@ -46,25 +46,29 @@ def parse_args():
         '--pulp-host',
         type=str,
         help='Pulp host',
-        required=True,
+        default=os.getenv('PULP_HOST'),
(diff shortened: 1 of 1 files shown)


scripts/notarize_build.py
```diff
@@ -0,0 +1,224 @@
...
+            )
+    return metadata
+
+
+async def main(
+    need_to_re_notarize: bool = False,
+):
+    args = parse_args()
+    pulp_client = PulpClient(
+        host=args.pulp_host,
```
javihernandez, line 133: Just a suggestion. How if we try resolving some of the configuration values from env variables? I'm mostly thinking in those used by albs-web-server. This way we can run it from a web_server container which, if not wrong, will have the env variables in place and we won't need to look for credentials, copy/paste ...

PR #1248, opened 2026-07-28
Match Pulp entities exactly when removing a product

### LINKED ISSUE #506: Fix product removal for products that start with the same name
Due to our mapping and incorrect determination of product entities, we can accidentally remove Pulp repositories/distributions that start with the same name

anfimovdm:
## Summary

Fixes AlmaLinux/build-system#506 — product removal could delete Pulp entities belonging to other products.

`remove_product()` fetched distributions with a `name__startswith` search on `pulp_base_distro_name` (`{owner}-{product}`) and then deleted every href in the result. That prefix also matches every product whose name starts with the same string, so removing `user/almalinux` deleted the distributions of `user/almalinux-extras` too.

While fixing that, two related problems in the same block:

- The guard meant to skip repositories that are absent from Pulp compared a **repository** name against a **distribution** name. Pulp names a distribution `{repo_name}-distro` (`PulpClient.create_rpm_distro`), so the comparison never matched, `continue` always fired, and the product's Pulp *repositories* were never deleted — only its distributions.
- The sign key repository of a community product (`{owner}-{product}-sign-key-repo`) is a Pulp *file* repository. Only the RPM distribution endpoint was queried, so it and its distribution were left behind on every removal.

## Changes

- `alws/crud/products.py` — `remove_product()` builds an exact `{repo.name}-distro` → repository mapping from `db_product.repositories` and filters the prefix-search result through it. The RPM and file distribution endpoints are queried concurrently, and each repository is deleted together with its distribution, so the "in DB but not in Pulp" case is still handled — now via the distribution's actual existence instead of a comparison that could never be true.
- `alws/utils/pulp_client.py` — new `get_file_distros()`, the file-distribution counterpart of `get_rpm_distros()`, using the same paginated `__get_entities` path.
- `tests/test_api/test_products.py` — regression test: simulates both distribution endpoints, asserts every product repository and distribution (including the sign key file repo) is deleted, and that a similarly-named product's distributions in both endpoints are untouched.
- `tests/fixtures/pulp.py` — new `get_file_distros` fixture; `test_user_product_remove` needs it now that removal queries that endpoint.

## Testing

`./run-tests.sh` — 85 passed, 10 skipped.

The new test was verified to fail against the unfixed code:

```
>       assert foreign_distro["pulp_href"] not in deleted_hrefs
E       AssertionError: assert '/pulp/api/v3/distributions/rpm/rpm/foreign/' not in [...]
```

That run also showed the second bug directly — only distribution hrefs were collected, no repository hrefs.

### DIFF
--- a/alws/crud/products.py
+++ b/alws/crud/products.py
@@ -305,44 +305,63 @@ async def remove_product(
         settings.pulp_password,
     )
     delete_tasks = []
-    # Pulp names a distribution after its repository, see
-    # PulpClient.create_rpm_distro() and PulpClient.create_file_distro()
-    repo_by_distro_name = {
-        f"{repo.name}-distro": repo for repo in db_product.repositories
-    }
-    # RPM repositories have an RPM distribution, the sign key repository
-    # of a community product is a file repository, so both endpoints
-    # have to be checked
-    distro_name_prefix = f"{db_product.pulp_base_distro_name}-"
-    found_distros = await asyncio.gather(
-        pulp_client.get_rpm_distros(
+    product_repo_names = {repo.name for repo in db_product.repositories}
+    # every pulp entity of a product is named after "{owner}-{product}",
+    # but so are the entities of a product whose name starts with the same
+    # string (e.g. "user-almalinux" and "user-almalinux-extras"), that's why
+    # the search results are matched against the product repositories
+    # by their exact name below
+    name_prefix = f"{db_product.pulp_base_distro_name}-"
+    # a community product keeps its sign key in a file repository,
+    # the rest of the product repositories are RPM ones
+    (
+        rpm_repos,
+        file_repos,
+        rpm_distros,
+        file_distros,
+    ) = await asyncio.gather(
+        pulp_client.get_rpm_repositories(
             include_fields=["pulp_href", "name"],
-            **{"name__startswith": distro_name_prefix},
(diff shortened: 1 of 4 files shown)


anfimovdm:
Follow-up after testing on a real Pulp: the `src` repository survived removal.

The first version still identified a distribution by its name (`{repo_name}-distro`) and a repository by the `pulp_href` stored in our DB. Both are assumptions rather than facts — a distribution whose name doesn't follow the convention is never matched, and a repository our DB has a stale href for is never found.

Removal now asks Pulp instead:

- repositories and distributions are listed from all four endpoints (`repositories/rpm/rpm`, `repositories/file/file`, `distributions/rpm/rpm`, `distributions/file/file`) with the `{owner}-{product}-` prefix;
- a repository is selected when its name is exactly one of the product's repository names, and it is deleted by the href Pulp reports, not the one in our DB;
- a distribution is selected when the `repository` it serves is one of those, with the `{repo_name}-distro` name as a fallback for distributions serving a publication rather than a repository.

`get_file_repositories()` was added to `PulpClient` alongside `get_file_distros()`.

The regression test now gives the `src` repository's distribution a name that breaks the convention and asserts it is still deleted, and adds a similarly-named foreign *repository* to both listings to confirm exact-name selection excludes it.

`./run-tests.sh` — 85 passed, 10 skipped.

